PR #8, opened 2023-07-19
[BBPBGLIB-1044] Sonata Replay

ferdonline:
## Context
BBPBGLIB-1044 Implement the ability of loading spikes for replay from a Sonata spike file.

## Scope
 - Implement the reader. For the moment use simply h5py
 - Integrate with Connection Manager

## Testing
 - Added a unit test for replay manager
 - Added an integration test


markovg:
Oh great! Thanks for implementing this!

neurodamus/replay.py
```diff
@@ -56,6 +59,15 @@ def open_spike_file(self, filename, delay):
 
         self._store_events(tvec, gidvec)
 
+    @classmethod
+    def _read_spikes_sonata(cls, filename, population):
+        spikes_file = h5py.File(filename, "r")
```
jdcourcol, line 64: Why not using libsonata and its reportreader ?

ferdonline: h5py Pros:
 - Our reader is 2 lines of code
 - We depend on h5py alone which we assume to be very stable
 - We can adapt quickly
 - One less API to know and monitor
Cons:
 - If the spec eventually changed we would need to change the code

Since IMO the risk of the spec changing is very low, in particular this one, I found reasonable to use h5py

matz-e: `h5py` with the transition to 3 was not very stable. We already depend on `libsonata`, and it would be good to use that as the source of truth. If the format changes, we have to deal with only one library to change and optimize.

ferdonline: Sure. What happened with h5py 3 transition btw?

matz-e: IIRC, they switched string handling, and/or `mpi4py` support. Led to a whole bunch of version pinning and incompatibilities in our deployment.

ferdonline: This is what we get with Sonata.
```
popa = spikes_file["NodeA"]
popa
<libsonata._libsonata.SpikePopulation object at 0x7fffd6946b30>
popa.get()
[(0, 0.10000000009999999), (2, 0.1500000001), (1, 0.17500000010000002), (0, 2.2750000000999986), (1, 3.0250000001000092), (2, 3.4500000001000153), (0, 4.350000000100028), (1, 5.700000000100047)...
```
The data structure is not compact at all and there seems to have happened some FP conversion problems.

I would like to get the raw arrays, numpy, no conversions. Is it possible?

mgeplf: > there seems to have happened some FP conversion problems.

Can you point me at the underlying file, so I can check that?

> I would like to get the raw arrays, numpy, no conversions. Is it possible?

I don't think that currently exists; if it's a requirement, I think opening a feature request would work.
The initial spike reader was implementation was provided by @alkino and worked on by @jorblancoa and @sergiorg-hpc.  Perhaps they can comment on how much work it would require to add the features you want.

ferdonline: It's a file generated in a test simulation: [sample_data/out.h5](https://github.com/BlueBrain/neurodamus/blob/BBPBGLIB-1044/sonata_replay/tests/sample_data/out.h5)

ferdonline: The conversion to a Python object seems to break it. Staying numpy looks ok
```
>>> f['/spikes/NodeA/timestamps'][0:10]
array([0.1  , 0.15 , 0.175, 2.275, 3.025, 3.45 , 4.35 , 5.7  , 6.975,
```

mgeplf: > The conversion to a Python object seems to break it, at least at display level. Staying numpy looks ok
> 
> ```
> >>> f['/spikes/NodeA/timestamps'][0:10]
> array([0.1  , 0.15 , 0.175, 2.275, 3.025, 3.45 , 4.35 , 5.7  , 6.975,
> ```

Are you sure it's not the numpy pretty printer?  What happens if you set `np.set_printoptions(precision=17)`?

ferdonline: Good one. Yep, now it looks more consistent
```
>>> np.set_printoptions(precision=17)
>>> f['/spikes/NodeA/timestamps'][0:10]
array([0.10000000009999999, 0.1500000001       , 0.17500000010000002, ...
```
Funny as `0.1000000000999` is not the closest possible representation of 0.1, but anyway I believe values should go 'as is".

ferdonline: Maybe @alkino or @jorblancoa could comment here the work required. However, since it's not available I would like to suggest we go as is and then I'm happy to switch to libsonata. Would you agree with that?

jorblancoa: I could add a couple of methods 'get_raw_node_ids()' and 'get_raw_timestamps()' to get the numpy arrays as in the file? What do you think @mgeplf ?

mgeplf: > What do you think @mgeplf ?

I think we should consider what it should look like a bit more before doing a "raw" sort of access.

For instance, having a function that:
1) does the filtering, as `get()` does now (ie: allow to give which ids we're interested in, and between which time span)
2) returns a tuple of numpy arrays: ie `(np.array(gids), np.array(times))`, which satisfy what @ferdonline needs, I think, and would allow filter and such to happen within the library, thus making it more performant for all the consumers (ie: bluepysnap, potentially brayns?)

Let's open an issue, and then get some feedback from people like @GianlucaFicarelli, @NadirRoGue and @joni-herttuainen

jorblancoa: I was trying to avoid changing the get() method, but maybe just a flag that by default is the current behaviour (list of tuples) and if the flag is set it return a tuple of 2 arrays as you suggested

mgeplf: I don't know if something on the C++ side would also be useful.
However, since it will be needed for the python version, it's probably worth starting on that, if you have the bandwidth.  We can decide on the exact names for functions after.

ferdonline: If that is open for discussion, I think a return type that could fit both purposes would be a numpy structured array so that we could access "line-by-line" as `array[0]`, access the property whole column as `array["timestamps"]`, sort together etc. IIRC pybind11 has an easy way to expose structs as record arrays.

ferdonline: @mgeplf From the side of other user applications, how does it sounds to replace the return type to the structured array? Are we too late to break the API?
Also, would you be ok with this PR going forward and adapting to libsonata later?

mgeplf: > So you prefer a new method?

I think that's a requirement for the c++.

> return type to the structured array?

I'm not sure structured arrays are the way to go since we don't use them in other places.  Have you check if they satisfy the performance expectations you have? The `std::vector` that is built is already just a `using Spike = std::pair<NodeID, double>;`, so I think you can just change the bindings to return that to test it.

jorblancoa: std::pair<NodeID, double> doesnt work with PYBIND11_NUMPY_DTYPE. I tried with a struct { NodeID node; double timestamp; } and that works fine, but the return type of the get() changes in both C++ (vector of Struct instead of vector of std::pair) and python (structured numpy array instead of list of tuples)
What i can do is to create another method in c++ that uses the struct instead of the pair, but i guess this would be duplicating the data...

ferdonline: > Have you check if they satisfy the performance expectations you have?

They would. They are compact in memory and even when we access `array["property"]` it uses a stride offset to access the data at once. For spikes this is more than enough.

> The std::vector that is built is already just a using `Spike = std::pair<NodeID, double>`;

Without knowing if such a change is too impactful, IMO turning that to a `Spike` struct looks actually nice.

mgeplf: > Without knowing if such a change is too impactful, IMO turning that to a Spike struct looks actually nice.

I think this is too large of an API change; Adding a function that returns multiple arrays or a struct, I think, would make more things cleaner from a breaking API perspective.

matz-e: There's a lot of buzz about Struct of Arrays and converting Arrays of Structs into them. Why not return a dictionary of numpy arrays? That would also easily convert into Pandas Dataframes for further processing.

mgeplf: > Why not return a dictionary of numpy arrays? That would also easily convert into Pandas Dataframes for further processing.

Good idea, I think that makes the most sense.  The pandas point is strong.

jorblancoa: So basically you mean to create an extra python only method (get_dict() for example) that just returns a python dict:
dict = get_dict()
dict["timestamps"]
dict["node_ids"]

jorblancoa: We should follow this discussions in the PR i created in libsonata: https://github.com/BlueBrain/libsonata/pull/285

I think this should be merged as it is until we agree on something there.

jorblancoa (approved):

neurodamus/replay.py
mgeplf: Does this handle the various sortings that the spike file can be written in? (ie: `These datasets may be unsorted, sorted by node id or sorted by spike time.`)

ferdonline: yes. Internally Neurodamus uses a multimap and sorts things per gid

ferdonline: We will revisit this reader later then for libsonata. Thanks all

PR #9, opened 2023-07-19
Dump load balance information on OutputRoot

iomaganaris:


iomaganaris:
This change partially fixes the issue. Now we have:
```
-bash-4.2$ ls /gpfs/bbp.cscs.ch/ssd/gitlab_map_jobs/bbpcihpcsim/P140508/hpc/sim/blueconfigs/sonataconf-quick-hip-multipopulation/output_coreneuron_sonataconf_lb_55aac866/sim_conf/_loadbal_a1d59a534f/
cx_hippocampus_neurons_most_central_100_SP_PC#.36.dat  cx_hippocampus_neurons_most_central_100_SP_PC#.dat
```
the circuit (?) specific data are generated in the output directory however the `mcomplex.dat` file is still generated in the current directory because of https://github.com/neuronsimulator/nrn/blob/d4b22d45d57b6d93b325384833db29bba45db55c/share/lib/hoc/loadbal.hoc#L572

ferdonline:
Hey @iomaganaris 

> I was wondering what are the access patterns of these files. ... how these files are written/read and what they are used for? Is this done from all the ranks of the simulation?
> we noticed a very large time for the calculation of the load balance and I was wondering whether this has anything to do with these files being in a different storage system than the rest of the outputs

I would find very strange that these files take any substantial time, even if all ranks were involved. That is because these are really lightweight txt files (`mcomplex.dat` might be a bit different as it's managed by Neuron). But anyway, we were careful there: when creating them in `_compute_save_complexities` only Rank0 writes. When reading `valid_load_distribution` operates only in rank0 and MPI broadcasts. The place where all ranks need the files from disk is when instantiating cells for simulation, but again, all quite lightweight

We may want to measure the independent phases. generating `mcomplex` should be fairly quick as well. IMO the big problem is with the load balance process, which requires instantiating all cells round-robin. This is huge compared to everything else, but hopefully will become lighter with the dry-run mode

iomaganaris:
> Hey @iomaganaris
> 
> > I was wondering what are the access patterns of these files. ... how these files are written/read and what they are used for? Is this done from all the ranks of the simulation?
> > we noticed a very large time for the calculation of the load balance and I was wondering whether this has anything to do with these files being in a different storage system than the rest of the outputs
> 
> I would find very strange that these files take any substantial time, even if all ranks were involved. That is because these are really lightweight txt files (`mcomplex.dat` might be a bit different as it's managed by Neuron). But anyway, we were careful there: when creating them in `_compute_save_complexities` only Rank0 writes. When reading `valid_load_distribution` operates only in rank0 and MPI broadcasts. The place where all ranks need the files from disk is when instantiating cells for simulation, but again, all quite lightweight
> 
> We may want to measure the independent phases. generating `mcomplex` should be fairly quick as well. IMO the big problem is with the load balance process, which requires instantiating all cells round-robin. This is huge compared to everything else, but hopefully will become lighter with the dry-run mode

Hi @ferdonline . I'm sorry but I think I was away on vacation and I never saw the reply.
Without looking into much details I think you're right regarding the performance aspects. Still it wouldn't be a bad idea to write this file in the output root instead of the current directory to avoid issues with reusing unintentionally the `mcomplex.dat` file when we want to clear the run directory.
If you think that's something we should pursue I can update the branch otherwise the PR can be closed.

iomaganaris:
We discussed this with Fernando on Slack. Ideally there should be some customization in the output directory of the load balance information files so this needs some more work and that's why it will stay as a draft PR atm

ferdonline:
Closing since this will not be tackled within the time of BBP

PR #15, opened 2023-07-25
[CNEUR-379] Use `/dev/shm` as a cache in multinode simulations

iomaganaris:
Currently in multinode simulations without `/dev/shm` all node write the `coreneuron_input` files in the same folder. This hurts a lot the `GPFS` performance for all users while this is happening.
Instead of all ranks writing to GPFS this PR adds a `CACHE` mode for `/dev/shm` where the `coreneuron_input` data are first staged to `/dev/shm` and then written by a single run in different folders in GPFS named `coreneuron_input_{node_id}`.
Then there `symlinks` are created from `coreneuron_input_{node_id}/*_{1,2,3}.dat` to `/dev/shm/.../coreneuron_input` and then `CoreNEURON` launches the simulation using  `/dev/shm/.../coreneuron_input`.
The only drawback of this approach is that there extra memory needed to use the `/dev/shm` cache since everything we dump to `/dev/shm` is saved in the RAM of the node

### DIFF
--- a/neurodamus/node.py
+++ b/neurodamus/node.py
@@ -10,7 +10,6 @@ import math
 import os
 import subprocess
 from os import path as ospath
-from pathlib import Path
 from collections import Counter, namedtuple, defaultdict
 from contextlib import contextmanager
 from shutil import copyfileobj, move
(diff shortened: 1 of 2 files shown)


neurodamus/node.py
```diff
@@ -1294,6 +1294,41 @@ def _sim_corenrn_write_config(self, corenrn_restore=False):
...
+            # performance. We can do that by finding out the Rank 0 of every node (already
+            # implemented).
+            # We will then create links on coreneuron_datadir for each file in the
+            # subfolders.
+            local_node_rank0 = int(SHMUtil.local_ranks[0])
+            if MPI.rank == local_node_rank0:
+                import shutil
+
+                node_specific_corenrn_output_in_storage = os.path.join(
+                    corenrn_output, "coreneuron_input_" + str(SHMUtil.node_id)
```
ferdonline, line 1314: Maybe we could turn ` corenrn_output` to a Path object and
```
node_specific_corenrn_output = corenrn_output / "coreneuron_input_" + str(SHMUtil.node_id)
...
os.symlink(node_specific_corenrn_output / filename, f)
```

iomaganaris: Good idea. I improved this a bit :+1:

neurodamus/node.py
```diff
@@ -1299,6 +1300,36 @@ def _sim_corenrn_write_config(self, corenrn_restore=False):
...
+            # implemented).
+            # We will then create links on coreneuron_datadir for each file in the
+            # subfolders.
+            local_node_rank0 = int(SHMUtil.local_ranks[0])
+            if MPI.rank == local_node_rank0:
+                import shutil
+
+                group_id = int(SHMUtil.node_id / 20)
+                node_specific_corenrn_output_in_storage = \
+                    Path(corenrn_output) / f"coreneuron_input/cycle_{self._cycle_i}/group_{group_id}/node_{SHMUtil.node_id}"
```
sergiorg-hpc, line 1321: **(Pasting from Slack)**

Thanks to a reminder by @1uc, I completely forgot that I discussed with 
@iomaganaris about generating the output in subfolders on GPFS, and that he already had something working recently 😅. Hence, I made some changes to improve it a bit by dividing the `coreneuron_input` directory into:
```
coreneuron_input/cycle_X/group_Y/node_Z 
```
Being `cycle_X` the current cycle in the model instantiation, `node_Z` the node ID (i.e., from 0 to 799 in the 800-node simulation), and `group_Y` a simple division to group the set of nodes in subfolders dividided by 20 (i.e., `group_id=floor(node_id / 20)`). Why 20? Well, a magic number just to divide the amount of folders into something reasonable.

With this simple approach, inside the `coreneuron_input` there would be **at most** 32 folders (i.e., one per cycle). Inside each cycle folder, there would be **at most** 40 subfolders (i.e., 800/20 = 40, one per subset of nodes). Inside the group of nodes, there would be **at most** 20 subfolders corresponding to the node IDs (i.e., from 0 to 19 in the first group, 20 to 39 in the second, and so on). Finally, inside the specific node folder, there would be **at most** 120 files from CoreNEURON (i.e., 3 files_per_rank x 40 ranks_per_node = 120).

In other words, we go from a folder with 3.1M files, to a tree that is much more manageable by GPFS and IME, regardless of the target file system that we use.

neurodamus/node.py
sergiorg-hpc: Sorry for nitpicking here, but we use `os.path` everywhere in Neurodamus and there is no actual benefit to swap this to `Path`. There are also some caveats of using `Path`, so let's not add more complexity than this code already has.

Check the new version, it looks quite clean "as is". If you guys still feel this must change for some reason, then let's discuss it and we change it everywhere and not just here.

neurodamus/node.py
sergiorg-hpc: Note that there has been a slight update in the code since this comment was written, but the reasoning is still valid.

neurodamus/node.py
```diff
@@ -1694,17 +1695,6 @@ def cleanup(self):
                     data_folder_shm = SHMUtil.get_datadir_shm(data_folder)
                     logging.info("Deleting intermediate SHM data in %s", data_folder_shm)
                     subprocess.call(['/bin/rm', '-rf', data_folder_shm])
-                    # Remove also the coreneuron_input_{node_id} folders
```
iomaganaris: Shouldn't we delete the symlinks in `/dev/shm` and folders in GPFS in the end?

iomaganaris:
If I'm not mistaken @sergiorg-hpc has been working in an improved version of this fix so this PR can now be closed?

iomaganaris:
We discussed about this change with Sergio offline and we concluded that it's something that it's not necessary for MMB simulations but might be beneficial so until needed this can be stayed open and Sergio can continue working on it if necessary

ferdonline:
Closing as it won't be tackled during BBP

PR #18, opened 2023-08-03
Update CoreConfig.mod for latest CoreNEURON changes

olupton:
- Parameter passed to solve_core should contain name of executable/process name as first argument in argv[].
 - This is required for changes in neuronsimulator/nrn/pull/2429.


alkino (approved):

pramodk (approved):
Ah! Forgot this !

thanks !

PR #21, opened 2023-08-09
Free stable identifiers listed for deletion.

1uc:
Also adds some printf debbuging statements for memory usage.

### DIFF
--- a/neurodamus/node.py
+++ b/neurodamus/node.py
@@ -37,8 +37,6 @@ from .utils.timeit import TimerManager, timeit
 # Internal Plugins
 from . import ngv as _ngv  # NOQA
 
-from neuron import h
-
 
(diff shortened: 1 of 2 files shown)


neurodamus/node.py
```diff
@@ -37,6 +37,8 @@
 # Internal Plugins
 from . import ngv as _ngv  # NOQA
 
+from neuron import h
```
ferdonline, line 40: Please do not import neuron directly which will start machinery too early. Use the `Neurodamus` (`Nd`) wrapper

1uc: Thank you, why is it fine to do so here:
https://github.com/BlueBrain/neurodamus/blob/8e388ba90dd196c03a6af63734fedab2f142c359/init.py#L9

PR #22, opened 2023-08-09
[BBPBGLIB-712] Estimate memory usage for synapse and connection 

st4rl3ss:
As detailed in [BBPBGLIB-712](https://bbpteam.epfl.ch/project/issues/browse/BBPBGLIB-712), this MR aims to introduce memory estimation for synapses and relative connections.

A basic counter is introduced that tracks all the synapses that get instantiated on all the ranks. Then memory usage is estimated using a LUT containing the approximate memory estimate for each type of synapse. The values were calculated using the script `synstat.py` included in the `_benchmark` folder.

Based on the counter and the memory estimate, an approximated figure for memory utilization can be reported to the user.


neurodamus/utils/memory.py
```diff
@@ -99,3 +99,27 @@ def get_mem_usage():
...
+            "Gap": 2.0,
+            "Glue": 0.5
+        }
+
+    def __getitem__(self, item):
+        return self._synapse_memory_usage[item]
+
+    def __setitem__(self, key, value):
+        raise ValueError("Cannot set values in SynapseMemoryUsage. "
+                         "Values are hardcoded.")
```
ferdonline, line 122: I would drop these two methods. If they are not implemented then `[]` operator is not enabled at all.

neurodamus/utils/memory.py
```diff
@@ -99,3 +99,27 @@ def get_mem_usage():
...
     return usage_mb
+
+
+class SynapseMemoryUsage:
+    ''' A small class that works as a lookup table
+    for the memory used by each type of synapse.
+    The values are in KB. The values cannot be set by the user.
+    '''
+    def __init__(self):
+        self._synapse_memory_usage = {
```
ferdonline, line 110: Since `_synapse_memory_usage` is constant I would promote to class_level. Then we can drop `__init__` and make `get_memory_usage` a classmethod to avoid requiring instances, which in this case makes little sense.

neurodamus/utils/memory.py
st4rl3ss: Makes sense. Removed them.

neurodamus/utils/memory.py
st4rl3ss: Thanks for the clarification. Fixed!

ferdonline:
Nitpick: In the benchmark script we could work in Kilobytes to avoid multiplying by 1024 bellow.
Apart from it LGTM.
IMO the only missing part here is that we output results to maybe a json file and not simply print stuff on the screen. But since that needs to be done for cell estimation maybe we can work that in a separate PR

st4rl3ss:
Thanks for the review @ferdonline . I think also we can integrate the json export into a different MR.
I will switch this MR to ready for review.

ferdonline (approved):

PR #23, opened 2023-08-10
Read 'electrodes_file' field from libsonata

jorblancoa:
Read file according to the spec: https://github.com/BlueBrain/sonata-extension/blob/master/source/sonata_tech.rst#format-of-the-electrodes_file

Addresses:
https://bbpteam.epfl.ch/project/issues/browse/BBPBGLIB-1038
https://bbpteam.epfl.ch/project/issues/browse/BBPBGLIB-924
https://bbpteam.epfl.ch/project/issues/browse/BBPBGLIB-1029

### DIFF
--- a/core/hoc/CompartmentMapping.hoc
+++ b/core/hoc/CompartmentMapping.hoc
@@ -91,8 +91,8 @@ func sectionNo() { local baseOffset, secIndex  localobj str, strobj
  * @param $o2 target object
  */
 
-proc registerMapping () { local cellIndex, activegid, nSoma, nAxon, nDend, nApic, nSec, nSeg, val, num_electrodes \
-  localobj gidvec, cellref, pyobj, commstr, target, cellDistributor, pc, python_obj
+proc registerMapping () { local cellIndex, activegid, nSoma, nAxon, nDend, nApic, nSec, indexSec, nSeg, val, num_electrodes \
+  localobj gidvec, cellref, pyobj, pop_info, commstr, target, cellDistributor, pc, python_obj
   strdef tstr
 
   cellDistributor = $o1
@@ -104,10 +104,12 @@ proc registerMapping () { local cellIndex, activegid, nSoma, nAxon, nDend, nApic
 
   // list of gids on current rank
   gidvec = cellDistributor.getGidListForProcessor()
+  pop_info = new PythonObject()
 
   for cellIndex=0, gidvec.size()-1 {
     activegid = gidvec.x[cellIndex]
     cellref = cellDistributor.getCell(activegid)
+    pop_info = cellDistributor.getPopulationInfo(activegid)
 
     // skip gids that we don't want to report, instead of skipping
     // we should add a flag in mapping
@@ -123,6 +125,7 @@ proc registerMapping () { local cellIndex, activegid, nSoma, nAxon, nDend, nApic
     nNode = 0
     nMyel = 0
     nSec = 0
+    indexSec = 0
 
     // vectors holding section and segments
     secvec = new Vector()
@@ -142,11 +145,11 @@ proc registerMapping () { local cellIndex, activegid, nSoma, nAxon, nDend, nApic
     python_obj = new PythonObject()
     has_lfp_manager = python_obj.hasattr(cellDistributor, "_lfp_manager")
     if( has_lfp_manager != 0 ) {
-        num_electrodes = cellDistributor._lfp_manager.get_number_electrodes(activegid)
-        lfp_factors = cellDistributor._lfp_manager.read_lfp_factors(activegid, secvec)
+        num_electrodes = cellDistributor._lfp_manager.get_number_electrodes(activegid, pop_info)
+        lfp_factors = cellDistributor._lfp_manager.read_lfp_factors(activegid, indexSec, nSoma, pop_info)
     }
     pc.nrnbbcore_register_mapping(activegid, "soma", secvec, segvec, lfp_factors, num_electrodes )
-
+    indexSec = indexSec + nSoma
     // gather axonal compartments and register with neuron
 
     nSec = 0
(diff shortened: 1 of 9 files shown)


docs/online-lfp.rst
```diff
@@ -0,0 +1,144 @@
...
+More information about this file can be found in the `SONATA Simulation Specification <https://github.com/BlueBrain/sonata-extension/blob/master/source/sonata_tech.rst#format-of-the-electrodes_file>`_
+
+Generating the Electrodes File (WIP)
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+The electrodes file can be generated using specific steps and considerations. Follow the instructions below to generate the file:
+
+1. Produce a compartment report from a target including the cells that will be used for the LFP calculation.
+
+2. Run the file getPositions.py, using the bash script GetPositions.sh. This will create a folder containing pickle files listing the (x,y,z) position of each segment in each cell in the target.
```
WeinaJi, line 21: Where are these scripts for generating the electrodes file? Will they be stored in neurodamus or in other repo?

neurodamus/cell_distributor.py
```diff
@@ -437,74 +437,84 @@ class LFPManager:
...
-                            scalar_factors.append(electrode_factor)
-            except KeyError:
-                logging.warning("GID %d not found in LFP file" % sonata_gid)
+                population_name, node_id = self.get_sonata_node_id(gid, population_info)
+                subset_data = self.get_node_id_subsets(node_id, int(section_index),
+                                                        int(num_sections), population_name)
+                for electrode_factors in subset_data:
+                    scalar_factors.append(Nd.Vector(electrode_factors))
+            except (KeyError, IndexError) as e:
+                logging.warning("Node id {} not found in the electrodes file: {}"
```
WeinaJi, line 502: Probably better to mention the `population_name` in the warning.

neurodamus/cell_distributor.py
```diff
@@ -437,74 +437,84 @@ class LFPManager:
...
-                sonata_gid = int(gid)
-                electrodes_group = self._lfp_file["electrodes"]["electrode_grid"][str(sonata_gid)]
-                num_electrodes = electrodes_group.shape[1]
-            except KeyError:
-                logging.warning("GID %d not found in LFP file" % sonata_gid)
+                population_name, node_id = self.get_sonata_node_id(gid, population_info)
+                subset_data = self.get_node_id_subsets(node_id, 0, 1, population_name)
+                num_electrodes = subset_data.shape[1]
+            except (KeyError, IndexError) as e:
+                logging.warning("Node id {} not found in the electrodes file: {}"
```
WeinaJi, line 516: The same, better to mention the population_name in the warning.

docs/online-lfp.rst
jorblancoa: Joseph has them in his public repo.
https://github.com/joseph-tharayil/create_lfp_weights_for_neurodamus

I told him to maybe create a bbp gitlab or bbp github for that but he doesnt have permissions. So i was planning to add the path to the repo once sorted out.

core/hoc/CompartmentMapping.hoc
```diff
@@ -104,10 +104,12 @@ proc registerMapping () { local cellIndex, activegid, nSoma, nAxon, nDend, nApic
 
   // list of gids on current rank
   gidvec = cellDistributor.getGidListForProcessor()
+  pop_info = new PythonObject()
 
   for cellIndex=0, gidvec.size()-1 {
     activegid = gidvec.x[cellIndex]
     cellref = cellDistributor.getCell(activegid)
+    pop_info = cellDistributor.getPopulationInfo(activegid)
```
ferdonline, line 112: Hopefully one day we pass the final cell_distributor, instead of the global one, so we don't need to get `pop_info`. But I agree that is another work and for the moment this is the best

core/hoc/CompartmentMapping.hoc
```diff
@@ -163,10 +166,10 @@ proc registerMapping () { local cellIndex, activegid, nSoma, nAxon, nDend, nApic
     }
 
     if( has_lfp_manager != 0 ) {
-        lfp_factors = cellDistributor._lfp_manager.read_lfp_factors(activegid, secvec)
+        lfp_factors = cellDistributor._lfp_manager.read_lfp_factors(activegid, indexSec, nAxon, pop_info)
```
ferdonline, line 169: Wouldnt be easier to get all lfp factors at once and then just pick as many we need in each step?

neurodamus/cell_distributor.py
```diff
@@ -437,74 +437,84 @@ class LFPManager:
...
+                                     "'scaling_factors', 'node_ids' and 'offsets' "
+                                     "in any of the populations {}.".format(population_list))
+
+    def get_sonata_node_id(self, gid, population_info):
+        return population_info[0], gid - population_info[1] - 1
+
+    def get_node_id_subsets(self, node_id, section_index, num_sections, population_name):
+        node_ids = self._lfp_file[population_name]["node_ids"]
+        # Look for the index of the node_id
+        index = numpy.where(numpy.array(node_ids) == node_id)[0][0]
```
ferdonline, line 470: `node_ids` is not a numpy array already? Also, If the array is sorted maybe we can use `searchsorted`

jorblancoa: node_ids is actually a <class 'h5py._hl.dataset.Dataset'> since it is coming from the h5 file.
We don't know for sure if the node_ids are sorted, since the electrodes_file is created from a report, and this report node_ids are not guaranteed to be sorted.

neurodamus/lfp_manager.py
```diff
@@ -0,0 +1,95 @@
...
+from .core import NeurodamusCore as Nd
+from .core.configuration import ConfigurationError
+
+
+class LFPManager:
+    """
+    Class handling the lfp functionality
+    """
+    def __init__(self):
+        self._lfp_file = None
```
ferdonline, line 13: Do you think it would be easy to make this a bit more RAII, so that, if we have `LFPManager` instance (not None) is always has the handle to the file? Smtg like
```
def __init__(self, lfp_weights_file):
    try:
        self._lfp_file = h5py.File(lfp_weights_file, 'r')
    except IOError as e:
        raise ConfigurationError(f"Error opening LFP electrodes file: {e}")
    self._available_populations = ...
```
That way we fail early if there's any problem, and the methods are simpler

ferdonline (commented):
Just left one comment, but nothing serious, feel free to ignore

neurodamus/lfp_manager.py
jorblancoa: Makes sense, however if we want to keep the construction of the LFPManager() in the GlobalCellManager constructor, we would need to pass around the path to the file and it doesnt feel right to me.

What i could do is to create the manager in the node.py, the same place where i call lfp_manager.load_lfp_config(lfp_weights_file, population_list)

But then it would fail at the same time than right now...

Otherwise we can merge this and see in the future

ferdonline (approved):

PR #27, opened 2023-08-14
Define the warning filter in pyutils and call it at the very beginning

WeinaJi:
## Context
"special" binaries which flushes denormal results to zero triggers the numpy warning for subnormal in every rank
after "import h5py". Neurodamus had a function to ignore this type of warning, but it is insufficient after the recent change of adding "import h5py" at the beginning of `replay.py` instead of in the function call. We are not seeing flooding warnings at the beginning of the simulation before instantiating neuron.

## Scope
Move the warning filter function in `pyutils.py` and call it in `__init__.py` before the first `import node` which calls `import h5py` from replay.py. In this way, we have to ignore the warnings in all ranks.

## Testing
No new test

## Review

### DIFF
--- a/neurodamus/__init__.py
+++ b/neurodamus/__init__.py
@@ -18,14 +18,14 @@ except Exception:
 __author__ = "Fernando Pereira <fernando.pereira@epfl.ch>"
 __copyright__ = "2018 Blue Brain Project, EPFL"
 
-# Control matched warning to display once in rank 0.
+# Ignore matched warning in all ranks.
 # "special" binaries built with %intel build_type=Release,RelWithDebInfo flushes
 # denormal results to zero, which triggers the numpy warning for subnormal in every rank
-# after "import h5py". Reduce this type of warning displayed once in rank0.
+# after "import h5py".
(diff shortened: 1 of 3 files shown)


WeinaJi:
@ferdonline , what do you think about this change? There are two options:
1. We move "import h5py" to the function call in "replay.py" and keep the filtering call not changed so it still works. 
2. We call the filtering in "\__init__.py" before importing node/neuordamus. But in this way we can only ignore the warning in all ranks instead of display once in rank0. That's because accessing the rank info triggers `_MPI._init_pc()` and `Neuron.load_hoc("netparmpi")` before  `_NeurodamusCore._init()` will skip the loading of hoc/mod as I saw from the failure of the base unit test.

WeinaJi:
> @ferdonline , what do you think about this change? There are two options:
> 
>     1. We move "import h5py" to the function call in "replay.py" and keep the filtering call not changed so it still works.

After sync with @ferdonline , we've decided to import h5py in the function call. So this PR is not needed.

PR #29, opened 2023-08-16
Print statistics of nodewise memory usage.

1uc:
Memory usage statistic of MPI tasks is useful when balancing work on a per task level. However, for estimating how close one is to going OOM, the min, max and average don't provide precise information. Instead it would be useful to see statistics of the memory usage of each compute node.

### DIFF
--- a/neurodamus/core/_shmutils.py
+++ b/neurodamus/core/_shmutils.py
@@ -1,6 +1,7 @@
 import os
 import psutil
 import subprocess
+import numpy as np
 
 class SHMUtil:
     """Helper class for the SHM file transfer mechanism of CoreNEURON.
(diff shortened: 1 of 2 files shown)


sergiorg-hpc (approved):

pramodk:
@1uc : for reference, could you just add an example output?

1uc:
Here's an example where one gets very different information depending on whether one looks at nodewise memory consumption or at an process level:
```
[INFO] (Aug.16 11:49:23) Memusage per node [MB]: Max=22999.98, Min=17308.74, Mean(Stdev)=17990.53(1671.74)
[INFO] (Aug.16 11:49:23) Memusage per task [MB]: Max=5591.37, Min=457.94, Mean(Stdev)=499.74(268.96)
```

PR #32, opened 2023-08-22
[BBPBGLIB-556] Full estimate of memory consumption

st4rl3ss:
Introduce full memory estimation capabilities for dry run as detailed in https://bbpteam.epfl.ch/project/issues/browse/BBPBGLIB-556.
When running in dry run mode, neurodamus will now provide a full estimate of the memory usage for both cells and synapses.
The synapse workflow is mostly untouched from the previous version with just a small bug fix.
For the cells estimate workflow we now perform full calculation of the estimate with the following workflow:

1. Finding all unique combinations of METypes and taking only the first 50 elements of each combination
2. Instantiating these combinations. Each set of maximum 50 elements is allocated in a different rank.
3. Getting the total memory consumption and then average it per cells.
4. Use the above average to calculate total memory usage for any combination with more than 50 elements.
5. Combine the cells memory total with the synapse one to get a full estimate.

The MR also introduces exporting the cells memory estimate to `json` file. Since this phase is the most time consuming one, by default the dry run workflow will automatically export the results for cell memory usage to a `memory_usage.json` file. It will also try to load this file on any subsequent run and just perform instantiation for METype combinations that are not already present in the file.

## To do/to check:

- [x] `etype` coming from `SONATA` files don't have any numerical suffix, while etypes taken from the `cells_info` internal object have a numerical suffix. This suffix is currently eliminated from any checks but should be clarified why they are different and if it's safe to eliminate the suffix.
- [x] check behaviour in more circuits or in case of multiple circuits
- [ ] add full dry run workflow test

## Review


ferdonline (dismissed):
I think this is a good start. A few points I think we should take care for follow-up work.
 - cell_distributor functions which dont handle dry_run should not see related vars
 - _instantiate_cells_dry is very specific to the `METype` detailed engine, and therefore, before we create a generic version it should go to `CellDistributor`
 - We are filtering cells to be instantiated with `_filter_memory_dict`. However later it may try to create connections for those cells and crash.
 - The display of Cell counts and memory taken could be a table
 - Take into account the projected synapse count (considering cell counts, not instantiated cells)

I gave a stab at improving some of these but I'd rather not block this and then we can move with follow-up PRs

neurodamus/node.py
```diff
@@ -502,7 +554,12 @@ def create_synapses(self):
             self._load_projections(pname, projection)
 
         if SimConfig.dry_run:
-            self._collect_display_syn_counts(synapse_counter)
+            self.syn_total_memory = self._collect_display_syn_counts(synapse_counter)
+            if MPI.rank == 0:
+                total_memory = self.cells_total_memory + self.syn_total_memory/1024
+                + self.avg_tasks_usage_mb*MPI.size
```
ferdonline, line 560: Maybe fix this one. The `+ self.avg_tasks_usage_mb*MPI.size` is an independent line, not part of the expression on line 559

st4rl3ss: Thanks for spotting this! I've fixed it.

st4rl3ss:
Thanks for the review @ferdonline!

Yeah, since this MR is getting very long and it's properly working already, we can leave any further development to another branch/MR. 

Having fixed the issue you signaled, If you have nothing against it, I would merge it when the test pass.

ferdonline (approved):

PR #41, opened 2023-09-04
[BBPBGLIB-1060] Remove CoreConfig.mod and enable CoreNEURON execution via NEURON Python API

iomaganaris:
## Context
Instead of using a mod file (`CoreConfig.mod`) to launch the CoreNEURON simulation using NEURON's C++ API we replace the file with NEURON's Python API to launch CoreNEURON in file mode.

## Scope
The changes in Neurodamus include:

- A new mechanism (mod file `CoreNEURONArtificialCell.mod`) to be instantiated in the processes that don't have any cells (i.e. simulation with 3 cells but 4 MPI ranks)
- New `CoreConfig` Singleton object that holds all the needed information for CoreNEURON and the related functions to write the simulation and report configurations read by CoreNEURON
- New tests
- Enabling again failing tests from https://github.com/BlueBrain/neurodamus/pull/38
- Disable `test_hoc_target_generate_subtargets` which fails in `python3-base` CI without any reason

Apart from the changes in Neurodamus there are also changes needed in NEURON which are done in https://github.com/neuronsimulator/nrn/tree/jblanco/coreneuron_from_neurodamus

## Testing
Unit tests for the new way to print the CoreNEURON configutarions were added.
All the other tests should pass without an issue apart from `py38-full-spack` which fails due to the dependence in the deployed module in BB5 and its mechanisms

## Review


iomaganaris:
Note that because of the changes in the mod files in this branch and the fact that `py38-full-spack` depends on the deployed modules, this test will always fail

iomaganaris:
With the latest changes and https://github.com/BlueBrain/spack/commit/f5e66bd6bd4ae54ce902d2268d1fd61bc223c991 this is now ready to merge.
I've run locally the tests from `py38-full-spack` and they now pass with the changes mentioned above

neurodamus/core/coreneuron_configuration.py
```diff
@@ -0,0 +1,117 @@
...
+        num_gids = len(gids)
+        logging.info(f"Adding report {report_name} for CoreNEURON with {num_gids} gids")
+        report_conf = f"{self.output_root}/{self.report_config_file}"
+        os.makedirs(os.path.dirname(report_conf), exist_ok=True)
+        with open(report_conf, "ab") as fp:
+            # Write the formatted string to the file
+            fp.write((f"{report_name} {target_name} {report_type} "
+                    f"{report_variable} {unit} {report_format} "
+                    f"{int(target_type)} {float(dt)} {float(start_time)} "
+                    f"{float(end_time)} {num_gids} {buffer_size}\n").encode())
```
matz-e, line 38: could we maybe have these one per line? Or
```
" ".join(map(str, [
    report_name,
    ...
    float(dt),
    ...
]))
```

neurodamus/core/coreneuron_configuration.py
```diff
@@ -0,0 +1,117 @@
...
+
+            # Write the array of integers to the file in binary format
+            fp.write(struct.pack(f'{num_gids}i', *gids))
+            fp.write(b'\n')
+
+    @run_only_rank0
+    def write_sim_config(
+            self, tstop, dt, forwardskip, prcellgid, celsius, v_init,
+            pattern=None, seed=None, model_stats=False, enable_reports=True):
+        simconf = f"{self.output_root}/{self.sim_config_file}"
```
matz-e, line 48: `pathlib.Path` would make all this a lot easier :)

neurodamus/core/coreneuron_configuration.py
```diff
@@ -0,0 +1,117 @@
+import os
+import logging
+from ._utils import run_only_rank0
+from . import NeurodamusCore as Nd
+
+class _CoreConfig(object):
```
matz-e, line 6: I find that name a wee bit ambiguous: the core of what? `CoreNeuronConfig` maybe?

neurodamus/node.py
```diff
@@ -270,6 +271,12 @@ def __init__(self, config_file, options=None):
             self._spike_populations = []
             Nd.execute("cvode = new CVode()")
             SimConfig.init(config_file, options)
+            if SimConfig.coreneuron:
+                CoreConfig.output_root = SimConfig.output_root
+                CoreConfig.datadir = SimConfig.coreneuron_datadir
+                # Instantiate the CoreNEURON artificial cell object which is used to fill up
+                # the empty ranks. This need to be done before the circuit is finitialized
+                CoreConfig.instantiate_artificial_cell()
```
matz-e, line 279: Could this be moved into `SimConfig.init`? Would maybe be a bit more logical

iomaganaris: `SimConfig` gets instantiated during validation of the configuration read by neurodamus and before MPI is initialized/imported. Since CoreConfig imports MPI we would like to keep it separated from SimConfig

neurodamus/core/coreneuron_configuration.py
```diff
@@ -0,0 +1,129 @@
...
+                report_format,
+                int(target_type),
+                float(dt),
+                float(start_time),
+                float(end_time),
+                num_gids,
+                buffer_size
+            ])) + "\n").encode())
+            # Write the array of integers to the file in binary format
+            fp.write(struct.pack(f'{num_gids}i', *gids))
```
ferdonline, line 52: Interesting. I thought `*gids` unpacking could be problematic. Imagine 1M gids translating to 1M arguments... I have tried locally and it seems python does it clever and doesnt ruin the stack

neurodamus/core/coreneuron_configuration.py
```diff
@@ -0,0 +1,129 @@
...
+        report_config.parent.mkdir(parents=True, exist_ok=True)
+        with report_config.open("a") as fp:
+            fp.write(filename)
+            fp.write("\n")
+
+    def psolve_core(self, save_path=None, restore_path=None):
+        from neuron import h
+        h.CVode().cache_efficient(1)
+        from neuron import coreneuron
+        from . import NeurodamusCore as Nd
```
ferdonline, line 115: In principle `Nd` is already a wrapper over `Neuron.h`, we can in principle avoid importing the latter

neurodamus/core/coreneuron_configuration.py
```diff
@@ -0,0 +1,129 @@
...
+                report_variable,
+                unit,
+                report_format,
+                int(target_type),
+                float(dt),
+                float(start_time),
+                float(end_time),
+                num_gids,
+                buffer_size
+            ])) + "\n").encode())
```
ferdonline, line 50: Why dont we use the same formatting as in the source? I'd find easier to maintain and likely more compatible
```suggestion
            fp.write("%s %s %s %s %s %s %d %lf %lf %lf %d %d\n" % (
                report_name,
                target_name,
                report_type,
                report_variable,
                unit,
                report_format,
                target_type,
                dt,
                start_time,
                floatend_time,
                num_gids,
                buffer_size,
            )
```

core/mod/CoreNEURONArtificialCell.mod
```diff
@@ -0,0 +1,18 @@
+COMMENT
```
ferdonline, line 1: For my understanding only, we have in place a way to create fake cells for CoreNeuron. Is this replacing it? If so, do you think we will be able to remove that Fake Cell workaround?

neurodamus/core/configuration.py
```diff
@@ -332,7 +332,7 @@ def _init_hoc_config_objs(cls):
             logging.info("User-defined RNG base seed %s", parsed_run.valueOf("BaseSeed"))
 
         if cls._simconf.coreNeuronUsed():
-            cls.coreneuron = h.CoreConfig(cls.output_root)
+            cls.coreneuron = True
```
ferdonline, line 335: Per comment above, if we dont want the var to hold the object maybe we we can delete it

neurodamus/core/configuration.py
```diff
@@ -209,7 +209,7 @@ class _SimConfig(object):
     _blueconfig = None  # new python BlueConfig parser
     _simconf = None
     rng_info = None
-    coreneuron = None  # bridge to CoreNeuron, instance of CoreConfig
+    coreneuron = None  # True if CoreNEURON is enabled
```
ferdonline, line 212: We already have a variable with this purpose. See `use_coreneuron`

neurodamus/node.py
```diff
@@ -1395,15 +1399,10 @@ def _run_neuron(self):
...
-        core_nrn_opts = ()
-
-        for opt, core_opt in {"save": "--checkpoint", "restore": "--restore"}.items():
-            opt_value = getattr(SimConfig, opt)
-            if opt_value is not None:
-                core_nrn_opts = core_nrn_opts + (core_opt, opt_value)
-
-        log_verbose("solve_core(..., %s)", ", ".join(core_nrn_opts))
-        SimConfig.coreneuron.psolve_core(*core_nrn_opts)
+        CoreConfig.psolve_core(
```
ferdonline, line 1402: Beautiful! 😄

neurodamus/core/coreneuron_configuration.py
ferdonline: Sorry, the format string should be a byte string, i.e. `b"%s %s. ..."`

core/mod/CoreNEURONArtificialCell.mod
iomaganaris: Unfortunately we can't replace the workaround since we need to have a cell (`ARTIFICIAL_CELL`) in every process. Then `ARTIFICIAL_CELL` needs to have a `NET_RECEIVE` and a `NET_RECEIVE` has to have a `net_event` so we still need this code

neurodamus/node.py
```diff
@@ -270,6 +271,12 @@ def __init__(self, config_file, options=None):
             self._spike_populations = []
             Nd.execute("cvode = new CVode()")
             SimConfig.init(config_file, options)
+            if SimConfig.use_coreneuron:
+                CoreConfig.output_root = SimConfig.output_root
+                CoreConfig.datadir = SimConfig.coreneuron_datadir
+                # Instantiate the CoreNEURON artificial cell object which is used to fill up
+                # the empty ranks. This need to be done before the circuit is finitialized
+                CoreConfig.instantiate_artificial_cell()
```
pramodk, line 279: @iomaganaris : just have this one question: could you remind me a rational for this? Because, I thought neurodamus/neuron handles the situation when some of the ranks don't have cells i.e. empty?

iomaganaris: I think that's the way Neurodamus handles ranks that don't have cells. It initializes an artificial cell for these ranks using the above mechanism. If we don't have this then we get an error saying that the rank doesn't have gids/cells assigned. (I don't remember now exact error but I can try and reproduce this is you want)
You can see also: https://github.com/BlueBrain/neurodamus/commit/1f12dd6e7ee32092f3b16e9ae27d883a07b2f799 and https://github.com/BlueBrain/neurodamus/commit/a1f6b9085e03bd79d0221349a8965c341f505d05

pramodk: Ok, thanks for info. Lets forget it then!

jorblancoa (approved):

PR #43, opened 2023-09-12
BBPBGLIB-1061 SONATA conf: Dont map to internal connectivity

ferdonline:
## Context

Address BBPBGLIB-1061 partially, ensuring, for the moment that we instantiate all projections.

## Scope
 - Consider only some specifically-named edge populations as internal connectivity
 - Other populations shall go as projections

Not all populations should go as projections since circuits still respect order, which is important for the case of e.g. NGV.

## Testing
Existing CI tests

## Review

### DIFF
--- a/neurodamus/connection_manager.py
+++ b/neurodamus/connection_manager.py
@@ -322,7 +322,7 @@ class ConnectionManagerBase(object):
 
     # - override if needed
     def _open_synapse_file(self, synapse_file, pop_name, n_nrn_files=None):
-        logging.info("Opening Synapse file %s, population: %s", synapse_file, pop_name)
+        logging.debug("Opening Synapse file %s, population: %s", synapse_file, pop_name)
         return self.SynapseReader.create(
             synapse_file, self.CONNECTIONS_TYPE, pop_name,
             n_nrn_files, self._raw_gids,  # Used eventually by NRN reader
--- a/neurodamus/io/sonata_config.py
+++ b/neurodamus/io/sonata_config.py
@@ -241,17 +241,6 @@ class SonataConfig:
                     circuit_conf["MorphologyType"] = "h5"
             circuit_conf["Engine"] = "NGV" if node_prop.type == "astrocyte" else "METype"
 
-            # find inner connectivity
-            for edge_config in network.get("edges") or []:
-                for edge_pop_name in edge_config["populations"].keys():
-                    edge_storage = self.circuits.edge_population(edge_pop_name)
-                    edge_type = self.circuits.edge_population_properties(edge_pop_name).type
-                    if edge_storage.source == edge_storage.target == node_pop_name and \
-                            edge_type == "chemical":
-                        edges_file = edge_config["edges_file"]
-                        if not os.path.isabs(edges_file):
-                            edges_file = os.path.join(os.path.dirname(self.network), edges_file)
-                        circuit_conf["nrnPath"] = edges_file + ":" + edge_pop_name
             return circuit_conf
 
         return {
@@ -263,6 +252,10 @@ class SonataConfig:
 
     @property
     def parsedProjections(self):
+        """
+        Build the projection blocks from configuration.
+        In SONATA all connectivity is handled as projections, not using internal connectivity.
+        """
         projection_type_convert = dict(
             chemical="Synaptic",
             electrical="GapJunction",
@@ -275,9 +268,8 @@ class SonataConfig:
             for population_name, edge_pop_config in edge_config["populations"].items():
                 edge_pop = self.circuits.edge_population(population_name)
                 pop_type = edge_pop_config.get("type", "chemical")
-                # skip unhandled synapse type or inner connectivity
-                if pop_type not in projection_type_convert or \
-                        (edge_pop.source == edge_pop.target and pop_type == "chemical"):
+                # skip unhandled synapse type
+                if pop_type not in projection_type_convert:
                     logging.warning("Unhandled synapse type: " + pop_type)
                     continue
                 edges_file = edge_config["edges_file"]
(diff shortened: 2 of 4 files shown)


jorblancoa (dismissed):

neurodamus/connection_manager.py
```diff
@@ -322,7 +322,7 @@ def open_synapse_file(self, synapse_file, edge_population, n_files=1, load_offse
 
     # - override if needed
     def _open_synapse_file(self, synapse_file, pop_name, n_nrn_files=None):
-        logging.info("Opening Synapse file %s, population: %s", synapse_file, pop_name)
+        logging.debug("Opening Synapse file %s, population: %s", synapse_file, pop_name)
```
WeinaJi, line 325: Why `debug`? I've found this info msg very helpful when I want to know the edges files that neurodamus loads without checking `circuit_config.json`.

ferdonline: I believe very similar information is printed anyway from another side, hence I decreased this specific line to debug. We can still still see it when debugging with `--debug`

matz-e: Who would print that? I agree with @WeinaJi that this is important information that should be in the logs most of the time, and not buried in debug output. There are other things that are less important in the default logs, IIRC.

ferdonline:
@matz-e there's doubt that's useful information, but if you look of a log produced with such run we have
```
[STEP] Circuit NodeB
[INFO] Init SynapseRuleManager. Options: {'load_offsets': False}
[INFO] Loading connections to population: <ConnectionSet: 1761-1761 (NodeB->NodeB)>
[INFO] No matching Connection blocks. Loading all synapses...

|-------20%-------40%-------60%-------80%------100%| [Time taken: 0.00 s]
[INFO]  * [ALL]. Created 2 connections
[STEP] Handling projections...
[INFO] Processing Edge file: /home/runner/work/neurodamus/neurodamus/tests/simulations/usecase3/edges_AB.h5
[INFO]  * NodeA__NodeB__chemical__NodeA-NodeB (Type: Synaptic, Src: NodeA, Dst: NodeB)
[INFO] Loading connections to population: <ConnectionSet: 4058-1761 (NodeA->NodeB)>
```
So for projections already print all information nicely. True I overlooked internal connectivity, which I was trying to get rid of in the first place 🤔
Since we will keep it for the moment, then I guess we are willing to have the duplicated info for projections.

neurodamus/connection_manager.py
ferdonline: With new patch:
```
[STEP] Circuit NodeB
[INFO] Processing edge file /home/runner/work/neurodamus/neurodamus/tests/simulations/usecase3/local_edges_B.h5, pop: NodeB__NodeB__chemical
```

ferdonline:
@jorblancoa @matz-e could you take another review pls?

neurodamus/io/sonata_config.py
```diff
@@ -39,6 +41,7 @@ def __init__(self, config_path):
         self._sim_conf = libsonata.SimulationConfig.from_file(config_path)
         self._entries = {}
         self._sections = {}
+        self._internal_edge_pops = set()
```
matz-e, line 44: Why is this a set if somewhere else we say that order matters?

Do we need this as an additional attribute, too? It seems to me that `circuit_conf["nrnPath"]` has the same contents and could be queried instead?

neurodamus/io/sonata_config.py
```diff
@@ -203,7 +207,7 @@ def Conditions(self):
         conditions["randomize_Gaba_risetime"] = str(conditions["randomize_Gaba_risetime"])
         return {"Conditions": conditions}
 
-    @property
+    @cached_property
     def Circuit(self):
```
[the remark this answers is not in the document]
ferdonline: It does unfortunately. That's a key name in BlueConfig

neurodamus/io/sonata_config.py
ferdonline: This set is internal, to keep track which edge populations are already "exposed" as internal connectivity, so the remaining are exposed as projections. For this purpose we are better served as a set and no need to expose as a property.

neurodamus/io/sonata_config.py
```diff
@@ -56,6 +59,7 @@ def __init__(self, config_path):
 
         self.circuits = libsonata.CircuitConfig.from_file(self.network)
         self._circuit_networks = json.loads(self.circuits.expanded_json)["networks"]
+        logging.debug(self.Circuit)  # compute Circuits so that it also sets _internal_edge_pops
```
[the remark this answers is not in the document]
ferdonline: We would need to compute these properties in the `__init__` in the controlled order.. I can't recall why I didn't go that way, maybe it's now easier with the cached property. Let me see

neurodamus/io/sonata_config.py
matz-e: Yes, but could we not just construct this set on the fly before instantiating all these "projections"?

neurodamus/io/sonata_config.py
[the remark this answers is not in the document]
matz-e: We're trying to get away from that, though? Do these have to correspond 1-1 with BlueConfig keys?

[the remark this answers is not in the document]
ferdonline: We will eventually. This Sonata reader is basically a conversion layer, so we expose keys like those in BlueConfigs.

neurodamus/io/sonata_config.py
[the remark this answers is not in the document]
ferdonline: I missed I already do that in `__init__`. Let me make this cleaner then

neurodamus/io/sonata_config.py
ferdonline: We have `Circuit` and `Projection` properties which both consume edges. Hence this change needs a shared object

jorblancoa (approved):

PR #47, opened 2023-09-15
Share GluSynapseHelper.hoc with public users

WeinaJi:
## Context
Requested by external users for running with the open-sourced plastic circuit.

## Scope
File copied from `neurodamus-models/common/hoc`


andrisecker (approved):
looking good!

PR #48, opened 2023-09-19
Modernize ALU.

1uc:
**Context**: We'd like to make `non_owning_identifier` a non-trivial class. Because this will allow us to make it reference counted. The implication is that it'll be  important that it's initialized and destroyed properly.

**Problem**: ALU.mod contain C code which `malloc`s `data_handle`s, but `data_handle`s contain a `non_owning_identifier`. Hence, `data_handle`s must be initialized and destroyed properly.

**Twist**: When install NEURON via `pip install` then `nrnivmodl` will generate a C file `ALU.c` which means we can't use `new` and `delete[]`.

**Fix**: Closer inspection of ALU.mod shows that it already requires a C++ compiler when `NRN_MECHANISM_DATA_IS_SOA` is defined. Since this covers the cases in which we need a C++ compiler, we use this macro to conditionally either use `malloc`/`free` or `new`/`delete`.

### DIFF
--- a/core/mod/ALU.mod
+++ b/core/mod/ALU.mod
@@ -135,7 +135,7 @@ VERBATIM {
     INFOCAST;
     Info* info = (Info*)hoc_Emalloc(sizeof(Info)); hoc_malchk();
     info->psize_ = 10;
-#ifdef NRN_MECHANISM_DATA_IS_SOA
+#ifdef NRN_VERSION_GTEQ_9_0_0
     info->ptrs_ = new handle_to_double[info->psize_]; hoc_malchk();
 #else
     info->ptrs_ = (handle_to_double*)hoc_Ecalloc(info->psize_, sizeof(handle_to_double)); hoc_malchk();
@@ -160,7 +160,7 @@ VERBATIM {
 #ifndef CORENEURON_BUILD
     INFOCAST;
     Info* info = *ip;
-#ifdef NRN_MECHANISM_DATA_IS_SOA
+#ifdef NRN_VERSION_GTEQ_9_0_0
     delete[] info->ptrs_;
 #else
     free(info->ptrs_);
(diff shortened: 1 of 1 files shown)


matz-e (approved):

core/mod/ALU.mod
```diff
@@ -135,7 +135,11 @@ VERBATIM {
     INFOCAST;
     Info* info = (Info*)hoc_Emalloc(sizeof(Info)); hoc_malchk();
     info->psize_ = 10;
+#ifdef NRN_MECHANISM_DATA_IS_SOA
+    info->ptrs_ = new handle_to_double[info->psize_]; hoc_malchk();
+#else
     info->ptrs_ = (handle_to_double*)hoc_Ecalloc(info->psize_, sizeof(handle_to_double)); hoc_malchk();
```
pramodk, line 141: @1uc : could you remind me why we need to use NEURON’s handle_to_double? I.e. when I saw this in the morning, I wanted to say this mod file shouldn’t use neurons api for memory management.

pramodk:
> Twist: When install NEURON via pip install then nrnivmodl will generate a C file ALU.c which means we can't use new and delete[].

just mentioning: this is because 8.* is still default release. If one installs 9a or when we will release 9.0, C++ should be default code generation.

core/mod/ALU.mod
```diff
@@ -135,7 +135,11 @@ VERBATIM {
     INFOCAST;
     Info* info = (Info*)hoc_Emalloc(sizeof(Info)); hoc_malchk();
     info->psize_ = 10;
+#ifdef NRN_MECHANISM_DATA_IS_SOA
```
pramodk, line 138: Oh! just additional question: I am not up to date with current usage/recommendation feom Olli: is it better/possible to use NRN_VERSION (>=9) macros than NRN_MECHANISM_DATA_IS_SOA? Later seems bit low level and shoudln’t be exposed in user MOD files if possible ?

(I see that it exists in old code but I wonder if it’s required)

pramodk: > Personally, I like NRN_MECHANISM_DATA_IS_SOA because it's more expressive than NRN_VERSIO

As a developer, yes!  As end users/scientists, not at all :) So I believe eversion macros are better.

pramodk (approved):
LGTM

PR #53, opened 2023-09-28
protect numpy.concatenate empty tuple in get_local_gids

WeinaJi:
## Context
Seen in some simulations, when a NodeSet taget doesn't contain any cell by its (bad) definition, `get_local_gids` raises an error from `numpy.concatenate(tuple([]))`. This PR sets an protection for this case.

## Scope
In the definition of `NodesetTarget::get_local_gids`, return `numpy.empty(0)` if the array to be concatenated is empty

## Testing
A new unit test `test_nodeset_target.py`

## Review

### DIFF
--- a/tests/test_nodeset_target.py
+++ b/tests/test_nodeset_target.py
@@ -1,10 +1,9 @@
 import numpy
 
-from neurodamus.core.nodeset import NodeSet
-from neurodamus.target_manager import NodesetTarget
-
 
 def test_get_local_gids():
+    from neurodamus.core.nodeset import NodeSet
+    from neurodamus.target_manager import NodesetTarget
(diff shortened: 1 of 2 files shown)


pramodk (approved):

tests/test_nodeset_target.py
```diff
@@ -0,0 +1,18 @@
+import numpy
+
+from neurodamus.core.nodeset import NodeSet
+from neurodamus.target_manager import NodesetTarget
```
ferdonline, line 4: `python3-base` tests still don't run forked and we have had hard-to-debug issues around that. I would suggest we do these imports inside the test test function

WeinaJi: Sure, I've moved them.

ferdonline (approved):

PR #55, opened 2023-10-04
[BBPBGLIB-1067] Implement 'node_sets_file' resolution order

jorblancoa:
## Context
Address BBPBGLIB-1067. 
The circuit_config nodesets are considered the base set of nodesets; the values in the simulation_config are added to the possible nodesets, and overwrite any duplicates from the base set.

## Scope
To accomplish this the new libsonata NodeSets update() method was used.

## Testing
Added unit test for new functionality

## Review


tests/test_nodeset.py
```diff
@@ -66,3 +67,54 @@ def test_ranges_overlap(ranges1, ranges2, expected):
...
+    # Create the test JSON files with nodesets
+    config_nodeset = tmpdir.join("node_sets_circuit.json")
+    with open(config_nodeset, 'w') as f:
+        json.dump(config_nodeset_file, f)
+
+    simulation_nodeset = tmpdir.join("node_sets_simulation.json")
+    with open(simulation_nodeset, 'w') as f:
+        json.dump(simulation_nodesets_file, f)
+
+    yield config_nodeset, simulation_nodeset
```
ferdonline, line 105: We could delete the files after they'r used, that's why generally we used `yield`s

neurodamus/target_manager.py
```diff
@@ -282,9 +284,17 @@ class NodeSetReader:
     Implements reading Sonata Nodesets
     """
 
-    def __init__(self, nodeset_file):
-        self.nodesets = libsonata.NodeSets.from_file(nodeset_file)
+    def __init__(self, config_nodeset_file, simulation_nodesets_file):
+        def _load_nodesets_from_file(nodeset_file):
+            if not nodeset_file:
+                return libsonata.NodeSets("{}")
```
ferdonline, line 290: Just to be sure, if the path is defined it must exist, otherwise it's an error?

neurodamus/io/sonata_config.py
```diff
@@ -166,6 +166,7 @@ def parsedRun(self):
         # "OutputRoot" and "SpikesFile" will be read from self._sim_conf.output
         # once libsonata resolves the manifest info
         parsed_run["OutputRoot"] = self._sim_conf.output.output_dir
+        parsed_run["TargetFileBase"] = self.circuits.node_sets_path
```
ferdonline, line 169: I find `TargetFileBase` a bit hard to understand. And maybe we don't even need a "BlueConfig"-like named entry

ferdonline (commented):
Couple of minor things

neurodamus/target_manager.py
jorblancoa: Yes, but it was the same before right?
```
self.nodesets = libsonata.NodeSets.from_file(nodeset_file)
```

neurodamus/io/sonata_config.py
jorblancoa: do you prefer something like parsed_run["config_nodeset_file"]? or parsed_run["circuits_node_sets_path"]?

ferdonline (dismissed):

ferdonline (approved):

PR #59, opened 2023-10-12
Add report dt to the logging

jorblancoa:
## Context
This will help us see when and if users set the report dt < 0.1 that when using CoreNEURON, it causes some performance degradation as reported in [REP-68](https://bbpteam.epfl.ch/project/issues/browse/REP-68)


pramodk (approved):
LGTM

ferdonline (approved):

PR #66, opened 2023-10-20
Remove misleading warning.

1uc:
The warning is misleading and causes users to run simulation with more than 10x too many resources. Since the magic cutoff is neither a lower bound, nor an upper bound (by many orders of magnitude), we should simply not print anything.

### DIFF
--- a/neurodamus/node.py
+++ b/neurodamus/node.py
@@ -372,12 +372,6 @@ class Node:
         target = self.target_manager.get_target(target_spec)
         cell_count = None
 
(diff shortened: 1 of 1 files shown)


pramodk (dismissed):
Since long time, I am in support of this :)

1uc:
There's two version, the milder make it an info; and the better option just don't say anything.6

ferdonline:
Thanks. I will close in favor of #67

PR #67, opened 2023-10-20
Use info instead of warning.

1uc:
See #66 for context. This version makes the misleading warning a hopefully less misleading info. The problem remains, that the magic threshold is off in both directions by orders of magnitude.


pramodk (approved):
If we want to still keep the advise, this looks fine to me as well!

ferdonline:
@1uc Could you please solve the style problem?
```
flake8: commands[0]> flake8
./neurodamus/node.py:377:30: F541 f-string is missing placeholders
```

ferdonline:
Closed after discussion with Luc

PR #69, opened 2023-10-25
BBPBGLIB-1076 Make all edge managers set load_offsets at init

ferdonline:
## Context
Even with a small execution we were hitting
```
323 NEURON: gid=1000000 already exists on this process as an output port
```
When outputting some debug it turned out that the post synaptic edge index was not being taken into account for the NGV netcon id. Therefore the id collision was almost certain.

## The problem

Turns out the `load_offset` property was not always set to True in the connection manager. 

This likely happens since in newer Sonata circuits there might be no internal connectivity, where such setting was being applied.

## Solution

With the current readers we always have synapse ids. So we could generalize the concept and set
 - All managers to load their synapse indices (ctor arg)
 - All except NGV which will never require them, so we explicitly disabled.
 - Setting applied even for connection managers holding projections only

## Testing

Tested with Katta's small NGV circuit.

## Review

### DIFF
--- a/neurodamus/connection_manager.py
+++ b/neurodamus/connection_manager.py
@@ -271,9 +271,9 @@ class ConnectionManagerBase(object):
         self._raw_gids = cell_manager.local_nodes.raw_gids()
         self._total_connections = 0
         self.circuit_conf = circuit_conf
-        self._load_offsets = False
         self._src_target_filter = None  # filter by src target in all_connect (E.g: GapJ)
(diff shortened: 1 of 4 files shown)


jorblancoa (approved):

cattabiani:
I could not reproduce the issue anymore. Thus, I consider this solved

cattabiani (approved):

PR #72, opened 2023-10-31
Manual container versions

heerener:
## Context
Containers automatically take a specific version of libsonata, libsonata-report and neuron. Sometimes a user wants to build with a specific version of one or more of these dependencies.

## Scope
* Allow setting the variables that determine which dependency version to use manually
* If set, also require a container version to be specified to avoid "polluting" the official container builds

## Review

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -23,6 +23,14 @@ variables:
   BLUECONFIGS_BRANCH:
     value: $BLUECONFIGS_BRANCH
     description: 'Name of the blueconfigs branch to test against'
+  LIBSONATAREPORT_TAG:
+    description: git tag for libsonata-report
+  LIBSONATA_TAG:
+    description: git tag for libsonata
+  NEURON_COMMIT_ID:
+    description: commit ID for neuron
+  REGISTRY_IMAGE_TAG:
+    description: this will be the container version - mandatory if you specify one of LIBSONATAREPORT_TAG, LIBSONATA_TAG, NEURON_COMMIT_ID
 
 python3-base:
   extends: .tox-template
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -23,14 +23,6 @@ variables:
   BLUECONFIGS_BRANCH:
     value: $BLUECONFIGS_BRANCH
     description: 'Name of the blueconfigs branch to test against'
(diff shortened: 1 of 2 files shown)


README.rst
```diff
@@ -85,6 +85,18 @@ Alternatively, you can start directly a neurodamus docker container where all th
 With the container, you can build your mod files and run simulations.
 See instructions in `docker/README.md <https://github.com/BlueBrain/neurodamus/blob/main/docker/README.md>`_.
 
+Building the docker container
+-----------------------------
+The docker container image is built automatically when a new tag is created in the repository, if a container with the specified version doesn't already exist.
+On manual pipeline runs, the container image is also built but not automatically pushed to docker-hub; you'll have to manually start the job in the pipeline you created if you want this to happen. Keep in mind that this is *optional*, the container image is *always* pushed to the gitlab registry!
```
iomaganaris, line 91: Could you please write where the CI needs to be launched from automatically? I still haven't found the CI run that build and publish the docker image. Today there was a new tag created. Did this trigger a special build in the CI that deployed the new image to DockerHub? If yes is there a link for it?

heerener: I've added some extra clarification - feel free to let me know if more is needed!

iomaganaris (approved):

.gitlab-ci.yml
```diff
@@ -80,8 +80,14 @@ configure-docker-image:
     - apt-get install -y git skopeo python3
     - |
       echo "Getting package version and checking whether we need to build the container"
-      PACKAGE_VERSION=$(git describe --tags)
-      REGISTRY_IMAGE_TAG=$(echo ${PACKAGE_VERSION%-*} | sed 's/-/.dev/')
+      if [ -n "${LIBSONATAREPORT_TAG}" || -n "${LIBSONATA_TAG}" || -n "${NEURON_COMMIT_ID}" && -z "${REGISTRY_IMAGE_TAG}" ]; then
```
WeinaJi, line 83: We need to define these 4 variables as the env `variables`, right?

heerener: You can pass any of them as variables when manually running the pipeline. If you do, you have to include `REGISTRY_IMAGE_TAG` (so you'll be specifying at least two of them :-) )

WeinaJi: ah, yes, indeed.

PR #75, opened 2023-11-01
Container fixes

heerener:
## Context
* It should be more clear which dependency versions were used for building a container
* At the time of tag creation, the spack recipes for dependencies may not be updated yet, and so a container would get built with the wrong dependency versions. 

## Scope
* Add dependency versions (libsonata-report tag, libsonata tag and neuron commit ID) as labels on the docker container
* Set container build jobs to run manually, even on tag creation. 

## Review


.gitlab-ci.yml
```diff
@@ -171,6 +175,7 @@ build-docker-image:
     - fi
   rules:
     - if: $CI_COMMIT_TAG
+      when: manual
```
ferdonline, line 178: Shouldnt this job always run if the preparation job (`configure-docker-image`) was triggered?

heerener: It should, but if I remove the `when` condition the pipeline will block (because it sees a job that is not getting finished). There may be a way around it, but I'll dig into that after (or during, if time permits) the retreat :-)

heerener: Turns out it had nothing to do with the `when: manual` condition, but rather with the combination of that with `rules`.

Demonstrates what happens when a job depends on a job that needs to be triggered manually.

When using only `when: manual` on a job and no `rules`, GitLab internally sets `allow_failure` to `true`. When using `rules` this doesn't seem to happen for some reason, as simply having a job in your pipeline with a rule that has `when: manual` configured will cause the pipeline to end with a `blocked` status. In order to replicate the desired behavior, we can manually add `allow_failure: true` to the rules that require the job to be started manually.

There's a small demo / playground / experimentation available here: https://bbpgitlab.epfl.ch/hpc/personal/heeren/cicd/-/blob/dependent-manual-job-with-rules/.gitlab-ci.yml?ref_type=tags

WeinaJi: That's tricky. I read somewhere that `allow_failure: true` is by default with `when: manual`. But I didn't know it changed inside `rules`.

WeinaJi (approved):

PR #80, opened 2023-11-13
[BBPP134-1022] Handling exceptions from NEURON during emodel loading

WeinaJi:
## Context
The current error handling between Neuron and neurodamus is as such:
* For a single process, Neuron (`hoc_execerror_mes`) throws a RuntimeError and neurodamus handles it (in `commands.py`) and aborts the process.
* For multiple processes, Neuron calls `nrnmpi_abort` when an error is raised in any rank. The exception can not be caught in neurodamus before the processes end.

As meaningful error msg is  requested in [BBPP134-1022], this PR sets Neuron not call `nrnmpi_abort` internally but passes the exception to be handled later in neurodamus.

## Scope
In `metype.py`, set the flag `old_flag = Nd.pc.mpiabort_on_error(0)` before the neuron process and revert it afterwards. When an exception from Neuron is caught by neurodamus, neurodamus will improve the error msg with more details and raise again.

## Testing
New unit test `test_error_handling.py`

## Review

### DIFF
--- a/neurodamus/metype.py
+++ b/neurodamus/metype.py
@@ -134,7 +134,8 @@ class METype(BaseCell):
         pass
 
     def __del__(self):
-        self._cellref.clear()  # cut cyclic reference
+        if self._cellref:
+            self._cellref.clear()  # cut cyclic reference
 
 
 class Cell_V6(METype):
@@ -163,8 +164,8 @@ class Cell_V6(METype):
             self._cellref = EModel(gid, morpho_path, morpho_file, *add_params)
             Nd.pc.mpiabort_on_error(old_flag)
         except Exception as e:
-            raise Exception("Error when loading Gid %d: emodel: %s, Morphology: %s"
-                            % (gid, emodel, morpho_file)) from e
+            raise RuntimeError("Error from NEURON when loading Gid %d: emodel: %s, Morphology: %s"
+                               ": %s" % (gid, emodel, morpho_file, str(e))) from e
         self._ccell = self._cellref
         self._synapses = Nd.List()
         self._syn_helper_list = Nd.List()
--- a/neurodamus/metype.py
+++ b/neurodamus/metype.py
@@ -156,7 +156,15 @@ class Cell_V6(METype):
         add_params = meinfos_v6.add_params or (keep_axon,)  # Keep axon incompatible with add_params
 
         logging.debug("Loading Gid %d: emodel: %s, Morphology: %s", gid, emodel, morpho_file)
-        self._cellref = EModel(gid, morpho_path, morpho_file, *add_params)
+        try:
+            # For this step, do not call mpi_abort in neuron and let neurodamus handle and abort,
+            # NB: Do not re-raise as ConfigurationError, neurodamus doesn't call mpi_abort so hangs
+            old_flag = Nd.pc.mpiabort_on_error(0)
+            self._cellref = EModel(gid, morpho_path, morpho_file, *add_params)
+            Nd.pc.mpiabort_on_error(old_flag)
+        except Exception as e:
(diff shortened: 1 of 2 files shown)


neurodamus/metype.py
```diff
@@ -156,7 +156,15 @@ def _instantiate_cell(self, gid, etype_path, emodel, morpho_path, meinfos_v6, de
         add_params = meinfos_v6.add_params or (keep_axon,)  # Keep axon incompatible with add_params
 
         logging.debug("Loading Gid %d: emodel: %s, Morphology: %s", gid, emodel, morpho_file)
-        self._cellref = EModel(gid, morpho_path, morpho_file, *add_params)
+        try:
+            # For this step, do not call mpi_abort in neuron and let neurodamus handle and abort,
+            # NB: Do not re-raise as ConfigurationError, neurodamus doesn't call mpi_abort so hangs
+            old_flag = Nd.pc.mpiabort_on_error(0)
```
jorblancoa, line 162: Shouldnt this be the default behaviour of neuron when called from Neurodamus?

neurodamus/metype.py
```diff
@@ -156,7 +156,15 @@ def _instantiate_cell(self, gid, etype_path, emodel, morpho_path, meinfos_v6, de
...
+            self._cellref = EModel(gid, morpho_path, morpho_file, *add_params)
+            Nd.pc.mpiabort_on_error(old_flag)
+        except Exception as e:
+            raise Exception("Error when loading Gid %d: emodel: %s, Morphology: %s"
+                            % (gid, emodel, morpho_file)) from e
```
1uc, line 167: Does neurodamus have an `Exception` class of its own, e.g. `NeurodamusError` or so?

neurodamus/metype.py
WeinaJi: Yes, I would like to start a discussion on it. 

I also think we should let neurodamus handle the exceptions from neuron. The only concern is that if letting neurodamus calls the mpi_abort, we will not have the stack traceback from neuron for debugging, is that acceptable?

neurodamus/metype.py
jorblancoa: It has ConfigurationError, but for config exceptions I guess..?

neurodamus/metype.py
1uc: My understand is that @WeinaJi added these lines after observing that it wasn't possible to catch the exception, because something had already called `MPI_Abort`.

I agree that this should be the default. I don't think it should be part of this PR, we can create an issue to "Review MPI_Abort during exception handling" or similar.

Weina, did you check that the above lines are correct, e.g. by inserting an exception that only happens on one MPI rank and observing that all MPI ranks are correctly killed? If yes, this, to me, looks good to merge.

neurodamus/metype.py
1uc: Then maybe `RuntimeError`?

neurodamus/metype.py
WeinaJi: I agree to decide whether using`Nd.pc.mpiabort_on_error(0)` as default in another PR, probably including more developers in the discussion. Let's focus on the emodel loading in this PR.

@1uc , yes, I've been testing with multiple ranks with 1 error cell. The exception is raised in one rank and neurodamus kills all the processes using the function `show_exception_abort`.

1uc: Very good. Then I'm out of idea what could go wrong.

1uc (dismissed):

neurodamus/metype.py
WeinaJi: The exception coming from neuron is of type `RuntimeError`, sure I can re-raise the same type.

jorblancoa (approved):

PR #81, opened 2023-11-16
Use default dt value of NEURON in WholeCell load balancing

iomaganaris:
Save the default `dt` from `NEURON` and use it when generating the `mcomplex` file with `ExperimentalMechComplex`

## Context
This PR solves the issue described in https://github.com/neuronsimulator/nrn/issues/2038

## Scope
First we save the original `dt` value of `NEURON` in the `SimConfig`. Then in the function that generates the complexity of each cell we set the `dt` to the original `NEURON` value, which normally should be `0.025`.

## Testing
Removed the explicit setting of `dt` to `0.025` in the load balance unit test

## Review

### DIFF
--- a/neurodamus/core/configuration.py
+++ b/neurodamus/core/configuration.py
@@ -213,6 +213,7 @@ class _SimConfig(object):
     # In principle not all vars need to be required as they'r set by the parameter functions
     blueconfig_dir = None
     current_dir = None
+    default_neuron_dt = 0.025
     buffer_time = 25
     save = None
     save_time = None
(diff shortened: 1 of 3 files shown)


neurodamus/core/configuration.py
```diff
@@ -626,7 +627,8 @@ def _global_parameters(config: _SimConfig, run_conf):
     h.celsius = config.celsius
     h.set_v_init(config.v_init)
     h.tstop = config.tstop
-    h.dt = run_conf.get("Dt", 0.025)
+    config.default_neuron_dt = h.dt
```
WeinaJi, line 630: Just to be sure that this change is equivalent: is `h.dt` already 0.025 by default?

iomaganaris: At the moment yes :+1:

WeinaJi (approved):

PR #84, opened 2023-11-22
Fix docs

heerener:
## Context
readthedocs builds were failing due to a new file that was not referred to anywhere.
Additionally, pre-merge test builds didn't show this.

## Scope
* Included online-lfp in the main table of contents
* Updated tox.ini with `-W -T` flags:
   * `-W` to turn warnings into errors
   * `-T` to show full traceback on exceptions

## Testing
n/a

## Review

### DIFF
--- a/docs/index.rst
+++ b/docs/index.rst
@@ -18,6 +18,7 @@ Contents
    Module Reference <api/modules>
    Developer Documentation <architecture>
    Developing a Custom Neurodamus <development>
+   Online LFP Calculation <online-lfp>
 
 
 Indices and tables
--- a/docs/index.rst
+++ b/docs/index.rst
@@ -18,7 +18,6 @@ Contents
    Module Reference <api/modules>
    Developer Documentation <architecture>
    Developing a Custom Neurodamus <development>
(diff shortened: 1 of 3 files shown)


WeinaJi:
@jorblancoa , do you think it good to link the online lfp doc in the main page ?

heerener:
> @jorblancoa , do you think it good to link the online lfp doc in the main page ?

If not, as an alternative we can try to make readthedocs less strict about warnings (and undo the tox change as well in that case).

jorblancoa:
I don't think is a good idea since the part to generate the weights file needs to be polished still...

heerener:
> I don't think is a good idea since the part to generate the weights file needs to be polished still...

It's apparently also possible to mark documents as orphans to avoid getting any warnings about them not being included anywhere. This seems like the best of both worlds: you can keep the warnings about unreferenced documents while at the same time you're not forced to include a document you don't think is ready as longs as you do it on purpose. 

I've changed the PR to mark `online-lfp` as an orphan. Don't forget to undo that once it's ready :-)

WeinaJi (approved):

PR #85, opened 2023-11-22
Read synapse parameters in a collective safe manner.

1uc:
## Context
When using `WholeCell` load-balancing, the access pattern when reading parameters during synapse creation is extremely poor and is the main reason why we see long (10+ minutes) periods of severe performance degradation of our parallel filesystem when running slightly larger simulations on BB5.

Using Darshan and several PoCs we established that the time required to read these parameters can be reduced by more than 8x and IOps can be reduced by over 1000x when using collective MPI-IO.  Moreover, the "waiters" where reduced substantially as well. See BBPBGLIB-1070.

Following those finding we concluded that neurodamus would need to use collective MPI-IO in the future.

We've implemented most of the required changes directly in libsonata allowing others to benefit from the same optimizations should the need arise. See,
https://github.com/BlueBrain/libsonata/pull/309
https://github.com/BlueBrain/libsonata/pull/307

and preparatory work:
https://github.com/BlueBrain/libsonata/pull/315
https://github.com/BlueBrain/libsonata/pull/314
https://github.com/BlueBrain/libsonata/pull/298 

By instrumenting two simulations (SSCX and reduced MMB) we concluded that neurodamus was almost collective. However, certain attributes where read in different order on different MPI ranks. Maybe due to salting hashes differently on different MPI ranks.

## Scope
This PR enables neurodamus to use collective IO for the simulation described above.

## Testing

We successfully ran the reduced MMB simulation, but since SSCX hasn't been converted to SONATA, we can't run that simulation.

## Review


neurodamus/io/synapse_reader.py
```diff
@@ -232,7 +232,8 @@ class SonataReader(SynapseReader):
     }
 
     def _open_file(self, src, population, _):
-        storage = libsonata.EdgeStorage(src)
+        hdf5_reader = libsonata.make_collective_reader(False, True)
```
WeinaJi, line 235: `libsonata.make_collective_reader(comm, collective_metadata, collective_transfer)`, maybe `comm` is missing ?

1uc: Yes, it used to have a default for `MPI.COMM_WORLD` where `MPI` is `from mpi4py import MPI`.

1uc: Thanks, changed.

neurodamus/io/synapse_reader.py
```diff
@@ -232,7 +232,13 @@ class SonataReader(SynapseReader):
     }
 
     def _open_file(self, src, population, _):
-        storage = libsonata.EdgeStorage(src)
+        try:
+            from mpi4py import MPI
+            hdf5_reader = libsonata.make_collective_reader(MPI.COMM_WORLD, False, True)
```
ferdonline, line 237: if `mpi4py` is installed for other reasons, but the user doesn't have libsonata_mpi, then either we have an unhandled AttributeError or make_collective_reader will return something non-collective. Am I missing something?

1uc: You avoid the former by requiring `libsonata>=0.1.25` (to be released) and the latter is intended because it prevents a hard dependency on MPI.

WeinaJi: I've just manually launched https://bbpgitlab.epfl.ch/hpc/sim/blueconfigs/-/pipelines/184571 with blueconfigs branch `weji/libsonata_mpi` and this neurodamus branch. The CI tests failed. Does it also require libsonata>=0.1.25 to be released ?

1uc: That looks like a deadlock, most like due to something being non-collective.

We "released" a compatible version of `libsonata` on BB5, so I don't think that's the issue. If it were we'd expect something related to not being able to import `libsonata.make_collective_reader`.

1uc: It's caused by this:
https://github.com/BlueBrain/HighFive/issues/911

WeinaJi:
Tested in pipeline https://bbpgitlab.epfl.ch/hpc/sim/blueconfigs/-/pipelines/189000 before merging

WeinaJi (approved):

PR #86, opened 2023-11-22
[BBPBGLIB-1092] Add unit tests for CLI options

jorblancoa:
## Context
Add unit tests to the CLI options in order to drop them from the blueconfigs CI.

## Review

### DIFF
--- a/tests/integration-e2e/test_cli_opts.py
+++ b/tests/integration-e2e/test_cli_opts.py
@@ -44,16 +44,15 @@ def test_save_restore_cli():
 
 
 def test_cli_prcellgid():
+    from neurodamus import Neurodamus
     test_folder = tempfile.TemporaryDirectory("cli-test-prcellgid")  # auto removed
     test_folder_path = Path(test_folder.name)
     for f in ("node_sets.json", "circuit_config.json", "input.dat", "simulation_config_mini.json"):
         shutil.copy(SIM_DIR / f, test_folder_path)
 
-    subprocess.run(
-        ["neurodamus", CONFIG_FILE_MINI, "--dump-cell-state", "62797"],
-        check=True,
-        cwd=test_folder_path
-    )
+    os.chdir(test_folder_path)
+    nd = Neurodamus(CONFIG_FILE_MINI, dump_cell_state=62797)
+    nd.run()
     assert (test_folder_path / "62798_py_Neuron_t0.0.nrndat").is_file()
     assert (test_folder_path / "62798_py_Neuron_t100.0.nrndat").is_file()
 
@@ -96,6 +95,7 @@ def test_cli_disable_reports():
     not os.environ.get("NEURODAMUS_NEOCORTEX_ROOT"),
     reason="Test requires loading a neocortex model to run")
 def test_cli_keep_build():
+    from neurodamus import Neurodamus
     with open(SIM_DIR / CONFIG_FILE_MINI, "r") as f:
         sim_config_data = json.load(f)
         sim_config_data["target_simulator"] = "CORENEURON"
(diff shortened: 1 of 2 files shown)


tests/test_cli_opts.py
```diff
@@ -63,3 +63,216 @@ def test_cli_prcellgid():
     )
     assert (test_folder_path / "62798_py_Neuron_t0.0.nrndat").is_file()
     assert (test_folder_path / "62798_py_Neuron_t100.0.nrndat").is_file()
+
+
+@pytest.mark.skipif(
+    not os.environ.get("NEURODAMUS_NEOCORTEX_ROOT"),
+    reason="Test requires loading a neocortex model to run")
```
ferdonline, line 70: I would love if we could not depend on neocortex so this could go into the new `integration` tests 😢

tests/test_cli_opts.py
```diff
@@ -63,3 +63,216 @@ def test_cli_prcellgid():
...
+        sim_config_data["output"]["output_dir"] = "output_keep_build"
+
+    test_folder = tempfile.TemporaryDirectory("cli-test-keep-build")  # auto removed
+    test_folder_path = Path(test_folder.name)
+    for f in ("node_sets.json", "circuit_config.json", "input.dat"):
+        shutil.copy(SIM_DIR / f, test_folder_path)
+    with open(test_folder_path / CONFIG_FILE_MINI, "w") as f:
+        json.dump(sim_config_data, f, indent=2)
+
+    subprocess.run(
```
ferdonline, line 118: Do you think we can tests without subprocess, but instead create a `Neurodamus` instance and running?

tests/test_cli_opts.py
jorblancoa: Me too...but I'm afraid is not possible... :S

tests/test_cli_opts.py
jorblancoa: Sure, I could try that. I just followed the format of the existing ones.

jorblancoa: I guess you mean something like:
`nd = Neurodamus(tmp_file.name, output_path=output_dir)`

And then instead of --output-path you test the option as a parameter to Neurodamus()?

tests/test_cli_opts.py
```diff
@@ -63,3 +63,216 @@ def test_cli_prcellgid():
...
+        capture_output=True,
+        text=True
+    )
+    assert "MULTI-CYCLE RUN: 2 Cycles" in result.stdout
+
+
+@pytest.mark.skipif(
+    not os.environ.get("NEURODAMUS_NEOCORTEX_ROOT"),
+    reason="Test requires loading a neocortex model to run")
+def test_cli_lb_mode():
```
WeinaJi, line 237: Same here with `test_loadbalance.py`. We need to reorg.

WeinaJi (commented):
Thanks for the adding the CLI tests. Is it possible to test `–enable-shm` option ? It is currently tested in blueconfigs.

WeinaJi:
> Thanks for the adding the CLI tests. Is it possible to test `–enable-shm` option ? It is currently tested in blueconfigs.

Forget about this comment. I overlooked the test , it is there in the PR.

tests/integration-e2e/test_cli_opts.py
```diff
@@ -44,15 +45,203 @@ def test_save_restore_cli():
...
+    os.chdir(test_folder_path)
+    nd = Neurodamus(CONFIG_FILE_MINI, dump_cell_state=62797)
+    nd.run()
+    assert (test_folder_path / "62798_py_Neuron_t0.0.nrndat").is_file()
+    assert (test_folder_path / "62798_py_Neuron_t100.0.nrndat").is_file()
+
+
+@pytest.mark.skipif(
+    not os.environ.get("NEURODAMUS_NEOCORTEX_ROOT"),
+    reason="Test requires loading a neocortex model to run")
```
ferdonline, line 63: These annotations are not needed anymore

WeinaJi (approved):

PR #87, opened 2023-11-23
Fix bug with duplicate count of synapses

ferdonline:
## Context
When using a different number of ranks it was noticeable a slightly different count in the number of synapses.

It turned out one was retrieving and counting synapses for the same gid under some circumstances when configure blocks would overlap.

## Scope
The bug is fixed by
 - keeping track of gids for which one has already retrieved synapse counts.
 - Exclude those when fetching and adding up the synapses counts per type to the global counter
 - Sort results order to improve comparison

## Testing
Running on a testing simulation 
 - 400 cells and 18 Configure blocks  (sonata-quick-v5-plasticity)
 - 1 vs 40 ranks (previously results would be different)

## Review


st4rl3ss (approved):
LGTM

PR #88, opened 2023-11-23
Implement chunking mechanism for loading synapses in dryrun mode

st4rl3ss:
## Context
Up until now for the purposes of counting in the dry run mode, all synapses were loaded at the same time. This started to create some issues when moving to very large circuit where the execution would consistently run out of memory.

This MR tries to fix this issue by chunking the loading in blocks of 100 cells at a time in order not to spend the whole memory.

## Scope
The changes are relatively small and include the adding of a small generator for chunking ranges in the util section, and a small rework of the loading synapses block to load cells in predefined ranges.

## Testing
No new functionality so no new tests are necessary.

## Review

### DIFF
--- a/neurodamus/connection_manager.py
+++ b/neurodamus/connection_manager.py
@@ -731,6 +731,7 @@ class ConnectionManagerBase(object):
           -  We will only consider gids which have not been accounted for yet.
         """
         BLOCK_LENGTH = 5000
+        SAMPLE_SIZE = 100
         raw_gids = dst_target.get_local_gids(raw_gids=True) if dst_target else self._raw_gids
         new_gids = numpy.setdiff1d(raw_gids, self._dry_run_counted_cells, assume_unique=True)
         if not len(new_gids):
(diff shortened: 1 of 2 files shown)


neurodamus/connection_manager.py
```diff
@@ -728,16 +730,32 @@ def _get_conn_stats(self, _src_target, dst_target):
...
+            sample_gids = new_gids[low:low + SAMPLE_SIZE]
+            sample_counts = self._synapse_reader.get_counts(sample_gids, group_by="syn_type_id")
+
+            extrapolated_counts = {
+                syn_type: count * (min(high, len(new_gids)) - low) / SAMPLE_SIZE
+                for syn_type, count in sample_counts.items()}
+            temp_counter.update(extrapolated_counts)
+            if cycle > 10:
+                log_all(VERBOSE_LOGLEVEL, "Rank: %d, %.2f%%", MPI.rank, (cycle+1)/n_blocks*100)
+        self._dry_run_counted_cells = numpy.union1d(self._dry_run_counted_cells, new_gids)
```
ferdonline, line 755: `_dry_run_counted_cells` contains the cells already considered. As it is we consider the whole range as counted, which I think is correct, but have an extra though about it.

st4rl3ss: I've extracted the extrapolation from the loop. Hope it makes the code simpler and more efficient.

st4rl3ss: I think it makes sense to flag them all as counted. After all we've already taken a sample from that block so, in the logic of not being able to count them all, they should be flagged as counted anyway.

Is there any particular reason why we shouldn't do so?

ferdonline: The only reason I could think about is, when handling "connection blocks" which overlap we would ensure some of their cells would be taken into account. But... it gets tricky to get the extrapolated counts right. We probably better trying to correctly sample over the whole range. It makes me think that we probably want to sample at regular intervals 🤔

ferdonline:
This approach is a good start. We can just maybe limit the number of buckets to 10 and then apply that change which samples from all metypes

ferdonline (dismissed):

ferdonline (approved):

PR #90, opened 2023-11-24
Remove shallow clone due to the recent setuptools

WeinaJi:
## Review


ferdonline:
I thought we would need to be deep enough only to get a tag. Instead of full clone, whatt about `depth: 50`?

heerener:
> I thought we would need to be deep enough only to get a tag. Instead of full clone, whatt about `depth: 50`?

I seem to remember that in the gitlab flow I got rid of the shallowness in order to avoid things suddenly breaking because there happened to be one or two more commits beyond the limit. I'm not sure the time invested into a "bump commit depth" fix is worth the efficiency gain from keeping the clone shallow.

WeinaJi:
> > I thought we would need to be deep enough only to get a tag. Instead of full clone, whatt about `depth: 50`?
> 
> I seem to remember that in the gitlab flow I got rid of the shallowness in order to avoid things suddenly breaking because there happened to be one or two more commits beyond the limit. I'm not sure the time invested into a "bump commit depth" fix is worth the efficiency gain from keeping the clone shallow.

Yes, I remember that fix. I was wondering how deep was enough and in the end I decided not to bother.  The shallow clone improves the time and space for large repos, like Neuron. In the future when neurodamus is big enough, we can review again.

PR #92, opened 2023-12-05
[BBPBGLIB-1027] Fix: Load Balance with multi-populations

ferdonline:
## Context

A number of Sonata circuits wouldn't use the load balance object, even though it had been built. See BBPBGLIB-1027

The issue stems from the fact that we compute load balance for a single population, but there was no way to identify which one. That information is required when later instantiating the circuit.

## Scope

This PR has two sides of it. 
 1. As an interim solution before we do load balance for all circuits, we pass the single load_balance object to all.
 2. An actual improvement to `LoadBalance`, making it aware of its population, so that
    - It won't mix and attempt creating a sub load balance out of another population
    - We can easily identify if the load-balancer is suitable for a given circuit

Additionally we _stopped_ enabling `MultiSplit` automatically since it's a pretty advanced and delicate option. Instead, when there are many more cores than cells, a warning is raised.

## Testing
```
neurodamus-py/tests/simulations/v5_sonata $ srun -Aproj16 -n2 neurodamus simulation_config.json --lb-mode=WholeCell --verbose
[...]
[STEP] LOADING NODES
[STEP] Circuit default
[VERB]  -> Nodes Format: NodeFormat.SONATA, Loader: load_sonata
[INFO] Reading Nodes (METype) info from '/gpfs/bbp.cscs.ch/project/proj1/circuits/SomatosensoryCxS1-v5.r0/O1-sonata/sonata/networks/nodes/default/nodes.h5'
[INFO]  => Cell distribution from Load Balance is valid
```

## Review

### DIFF
--- a/neurodamus/node.py
+++ b/neurodamus/node.py
@@ -383,16 +383,15 @@ class Node:
             circuit.CircuitPath if is_sonata_config
             else self._run_conf["nrnPath"] or circuit.CircuitPath
         )
-        load_balancer = LoadBalance(lb_mode, data_src, self._target_manager, prosp_hosts)
+        pop = target_spec.population
+        load_balancer = LoadBalance(lb_mode, data_src, pop, self._target_manager, prosp_hosts)
 
         if load_balancer.valid_load_distribution(target_spec):
             logging.info("Load Balancing done.")
             return load_balancer
 
         logging.info("Could not reuse load balance data. Doing a Full Load-Balance")
-        cell_dist = self._circuits.new_node_manager(
-            circuit, self._target_manager, self._run_conf
-        )
+        cell_dist = self._circuits.new_node_manager(circuit, self._target_manager, self._run_conf)
         with load_balancer.generate_load_balance(target_spec, cell_dist):
             # Instantiate a basic circuit to evaluate complexities
             cell_dist.finalize()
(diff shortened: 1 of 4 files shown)


neurodamus/node.py
```diff
@@ -455,6 +453,7 @@ def create_cells(self, load_balance=None):
                 logging.warning("Skipped node population (restrict_node_populations)")
                 continue
             self._circuits.new_node_manager(circuit, self._target_manager, self._run_conf,
+                                            load_balancer=load_balance,
```
WeinaJi, line 456: As I understand we run `compute_load_balance` on the first non-virtual population of the circuit, so shall the load_balancer be applied to that population?  
Is it feasible to compute lb for all populations and load them by population?

ferdonline: Actually in the currently implementation I'm curious whether it would be able to resume for circuits which loadbal is provided but doesnt contain info for that circuit

ferdonline: The last commit should solve the issue. I will start drafting applying loadbalance for all circuits

WeinaJi: The applying part looks good, thanks for implementing the population information along with the lb files. 
One more question, I see in `compute_load_balance` ( line 345 of the same file), we still compute LB for the first population. Is there a plan to compute for all?

ferdonline: Yes, that is a logical next step I'd like to tackle as I have time

WeinaJi (approved):

PR #93, opened 2023-12-05
[BBPBGLIB-958] Ensure node populations come from a single source

ferdonline:
## Context
Ensure node populations (Sonata) come from a single source.

Node populations are considered unique. Cases where several regions are composed of multiple populations, each must have a different name.

## Scope
Add checks in the Sonata config loader.

## Review


WeinaJi:
I see that libsonata already rejects the `circuit_config.json` when the same node population appears twice in the config file.
```
conf=CircuitConfig.from_file("circuit_config.json")
libsonata._libsonata.SonataError: Population NodeA is declared twice
```

ferdonline:
Thanks @WeinaJi. In that case I think we can close this

PR #99, opened 2023-12-08
Update documentation for new synapse estimate algorithm

st4rl3ss:
## Context
Simple MR to update docs to reflect the new mechanism used to estimate synapse counts.

## Scope
No code update, just better documentation.

## Testing
No testing needed, just a documentation update.

## Review
* N/A Unit/Scientific test added
* N/A Updated Readme, in-code, developer documentation


docs/architecture.rst
```diff
@@ -367,25 +367,32 @@ usage before and after the instantiation. The memory usage is then averaged over
...
+while keeping good performances.
+
+For each metype and block, we estimate the number of synapses by sampling a subset (100) of the gids
+and counting the number of synapses for each gid in the sample. The counts are then extrapolated
+linearly to the entire population of gids for that metype.
+
+We used a sophisticated sampling strategy to optimize performance. It processes the gids in blocks,
+starting with small blocks (5000) and gradually increasing the block size using a geometric progression.
+Within each block, it samples a fixed number of gids (always 100).
+This strategy takes advantage of data locality and allows us to handle very large sets of gids.
```
ferdonline, line 384: I think the idea is getting a bit repeated. We could maybe condense this and go straight to the point. Also `counts are then extrapolated linearly to the entire population` is not entirely true. They are extrapolated linearly only inside the block to work with skewed data as we have recently changed.
What about. ?
The parameters of the sampling are as follows:
 - Blocks start with length of 5000, increasing at a rate of 10%
 - Count synapses for the first 100 cells of the block (take advantage of data locality)
 - Extrapolate for the whole block and add to the global metype estimate.

neurodamus/connection_manager.py
```diff
@@ -725,15 +725,25 @@ def target_gids(gids):
 
     def _get_conn_stats(self, _src_target, dst_target):
         """Estimates the number of synapses per type for the given destination target
+
+        Args:
+            _src_target: not used
+            dst_target: The target to estimate synapses for
```
ferdonline, line 731: At this point we know that `_src_target` wont likely ever be used. I would drop it from the signature and modify the comment
```suggestion
    def _get_conn_stats(self, dst_target):
        """Estimates the number of synapses per type for the given destination target.
        Note: measurement is done without any restriction on the source target. As so
        it can avoid counting the same synapse twice by keeping track of target cells alone.

        Args:
            dst_target: The target to estimate synapses for
```

ferdonline:
Thanks for the docs. I left some suggestions for maybe adding a bit more clarity, but feel free to do differently

st4rl3ss:
> Thanks for the docs. I left some suggestions for maybe adding a bit more clarity, but feel free to do differently

Thanks a lot for the suggestions. They were all integrated.

ferdonline (approved):

PR #100, opened 2023-12-08
[BBPBGLIB-854] Implement a CLI option to run coreneuron in direct mode without writing model data to disk

WeinaJi:
## Context
This PR implements a new CLI option "--coreneuron-direct-mode" that enables to run the coreneuron simulation using direct memory transferred from neuron, without writing intermediate model data to disk.

## Scope

- When "--coreneuron-direct-mode", we set the Coreneuron parameter`coreneuron.file_mode=False`. 
- And Coreneuron writes the element reports but not the spike report.  So we still create `sim.conf`, `report.conf` and an empty folder `coreneuron_input`. The spike report is created explicitly in neurodamus via `sonata_spikes()` just like the Neuron runs.
- Enable fast membrane current calculation `Nd.cvode.use_fast_imem(1)` for  `i_membrane` and `lfp` reports.

## Testing
New test file `test_coreneuron_directmode.py`.
The blueconfig pipeline tests with  "--coreneuron-direct-mode" enabled have been done manually in https://bbpgitlab.epfl.ch/hpc/sim/blueconfigs/-/pipelines/219862 using the blueconfig [branch](https://bbpgitlab.epfl.ch/hpc/sim/blueconfigs/-/merge_requests/114).

## Review

### DIFF
--- a/neurodamus/commands.py
+++ b/neurodamus/commands.py
@@ -61,7 +61,8 @@ def neurodamus(args=None):
         --model-stats           Show model stats in CoreNEURON simulations [default: False]
         --dry-run               Dry-run simulation to estimate memory usage [default: False]
         --num-target-ranks=<number>  Number of ranks to target for dry-run load balancing
-        --skip-write-model      Run CoreNeuron in direct mode, without writing model data to disk
+        --coreneuron-direct-mode     Run CoreNeuron in direct memory mode transfered from Neuron,
+                                     without writing model data to disk.
     """
     options = docopt_sanitize(docopt(neurodamus.__doc__, args))
     config_file = options.pop("ConfigFile")
--- a/neurodamus/commands.py
+++ b/neurodamus/commands.py
@@ -61,6 +61,7 @@ def neurodamus(args=None):
         --model-stats           Show model stats in CoreNEURON simulations [default: False]
         --dry-run               Dry-run simulation to estimate memory usage [default: False]
         --num-target-ranks=<number>  Number of ranks to target for dry-run load balancing
+        --skip-write-model      Run CoreNeuron in direct mode, without writing model data to disk
     """
     options = docopt_sanitize(docopt(neurodamus.__doc__, args))
     config_file = options.pop("ConfigFile")
--- a/neurodamus/core/configuration.py
+++ b/neurodamus/core/configuration.py
@@ -1021,7 +1021,7 @@ def _coreneuron_direct_mode(config: _SimConfig, run_conf):
             logging.warning("--coreneuron-direct-mode not valid for multi-cyle model building, "
                             "continue with file mode")
             direct_mode = False
-        if config.save:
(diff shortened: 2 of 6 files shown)


neurodamus/commands.py
```diff
@@ -55,6 +55,7 @@ def neurodamus(args=None):
         --enable-shm=[ON, OFF]  Enables the use of /dev/shm for coreneuron_input [default: ON]
         --model-stats           Show model stats in CoreNEURON simulations [default: False]
         --dry-run               Dry-run simulation to estimate memory usage [default: False]
+        --skip-write-model      Run CoreNeuron in direct mode, without writing model data to disk
```
matz-e, line 58: I'm wondering if this would be better as a `--write-model` and `--no-write-model` pair of options? Would make the `if not skip_write_model` a bit friendlier as `if write_model`.

Otherwise great to see this!

tristan0x: There are several options _--do-that=[ON, OFF]_
Maybe this new option should be consistent with this pattern i.e "--write-model=[ON, OFF]"

WeinaJi: I've renamed the CLI option "--coreneuron-direct-mode".  Does it sound better?

neurodamus/core/configuration.py
```diff
@@ -1005,6 +1007,32 @@ def _spikes_sort_order(config: _SimConfig, run_conf):
...
 
 
+@SimConfig.validator
+def _coreneuron_direct_mode(config: _SimConfig, run_conf):
+    user_config = config.cli_options
+    direct_mode = user_config.coreneuron_direct_mode
+    if direct_mode:
+        if config.use_neuron:
+            logging.warning("--coreneuron-direct-mode not valid for NEURON, will do nothing")
+            direct_mode = False
```
ferdonline, line 1017: IMO we can and probably want to error out so the user can fix his inout. Discarding this CLI option because the Simulator is Neuron sounds dangerous, maybe the user really wanted CoreNeuron

WeinaJi: Ok, I've changed to raise an exception in this case.

jorblancoa (approved):

matz-e (commented):
LGTM

neurodamus/commands.py
```diff
@@ -61,6 +61,8 @@ def neurodamus(args=None):
         --model-stats           Show model stats in CoreNEURON simulations [default: False]
         --dry-run               Dry-run simulation to estimate memory usage [default: False]
         --num-target-ranks=<number>  Number of ranks to target for dry-run load balancing
+        --coreneuron-direct-mode     Run CoreNeuron in direct memory mode transfered from Neuron,
+                                     without writing model data to disk.
```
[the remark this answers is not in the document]
WeinaJi: Sure, we will review this CLI Option in another PR.

PR #101, opened 2023-12-14
Deprecate BlueConfig p1: Raise errors for BlueConfig configuration files

WeinaJi:
## Context
This PR starts the deprecation of BlueConfig configuration files.

## Scope
 As the first step, in this PR we raises an error while `--configFile` points to any non-json file.  By default, `--configFile` is changed to "simulation_config.json" instead of "BlueConfig".

## Testing
Migrate the BlueConfig files to SONATA configurations for all the current unit tests.

## Review


tests/integration-e2e/test_plugin.py
```diff
@@ -132,6 +133,8 @@ class ACellEngine(EngineBase):
     InnerConnectivityCls = ACellSynapseManager
 
 
+@pytest.mark.skip(
+    reason="Cannot test with SONATA configs, no SONATA parameter for Engine")
```
WeinaJi, line 137: @ferdonline  I disable this test because I cann't find a way to convert it to SONATA config.  Do you have any suggestion?

neurodamus/node.py
```diff
@@ -251,10 +251,14 @@ def __init__(self, config_file, options=None):
             config_file: A BlueConfig file
             options: A dictionary of run options typically coming from cmd line
         """
-        if config_file and config_file.endswith(".json"):
-            import libsonata
-            conf = libsonata.SimulationConfig.from_file(config_file)
-            Nd.init(log_filename=conf.output.log_file)
+        if config_file:
+            if config_file.endswith(".json"):
```
jorblancoa, line 255: Thanks for this PR! Super nice.

Just a small nitpick.
Since from now on we only support SONATA, maybe we should raise the ConfigurationError in case the config_file passed is "BlueConfig" and then leave the libsonata.SimulationConfig.from_file() function take care of raising an exception if the file is not correct, instead of checking for a ".json" file

tests/integration-e2e/test_plugin.py
ferdonline: You can leave it disabled for the moment. I will refactor this so that we can keep testing alternate engines even without Sonata syntax

neurodamus/node.py
WeinaJi: That's also an option. In this case, libsonata raises
```
RuntimeError: Path `BlueConfig` is not a file
```
Is this error msg self-explanatory ?

jorblancoa: But this is in the case of BlueConfig not existing right? what if it exists?

WeinaJi: Ah, you are right, I was testing in the wrong folder. Then in the case of existing BlueConfig 
```
RuntimeError: [json.exception.parse_error.101] parse error at line 1, column 1: syntax error while parsing value - invalid literal; last read: 'R'
```

jorblancoa: Exactly, thats why i thought that having a first if config_file == "BlueConfig" raise BlueConfig not supported etc
and then else let libsonata raise in case of wrong format

WeinaJi: I see I total misunderstood :laughing: Now I got it. Sure, I will make the change.

jorblancoa (approved):

PR #103, opened 2023-12-15
[BBPBGLIB-1093] Move scientific tests from blueconfigs

jorblancoa:
## Context
Moved scientific tests 'test_synapse_delay_override' and 'test_many_replay_sim' from blueconfigs.

## Review


neurodamus/connection.py
```diff
@@ -803,7 +803,14 @@ def create_on(self, conn, sec, syn_obj, syn_params):
             log_all(logging.DEBUG, "Creating Replay on %d-%d, times: %s",
                     conn.sgid, conn.tgid, self.time_vec.as_numpy() if self.has_data() else "N/A")
 
-        nc = Nd.NetCon(vecstim, syn_obj, 10, syn_params.delay, syn_params.weight, sec=sec)
+        nc = Nd.NetCon(
+            vecstim,
+            syn_obj,
+            10,
+            conn.syndelay_override or syn_params.delay,
```
WeinaJi, line 810: I implemented `syndelay_override` but I can't remember why we didn't add it here before. Isn't `_attach_source_cell` sufficient?

jorblancoa: probably @ferdonline knows more...but we were trying to see a change in the unit test and until we added that nothing was happening...

WeinaJi: maybe `_attach_source_cell` only influence the replay with coreneuron. Can you add the test with coreneuron as well?

WeinaJi (approved):

PR #104, opened 2023-12-18
Fix unit test: new morphology test file to fulfil MorphIO change

WeinaJi:
## Context
MorphIO with commit aabd346 is strict about soma types: ASC/H5 are contours, SWC is single point, cylinders or neuromorpho 3 point. One of the test morph files used by our unit tests is not valid after this change because of morphology contour with only a single point.

## Scope
New `tests/sample_data/morphology/simple.h5` is copied from MorphIO repo `MorphIO/tests/data/h5/v1/simple.h5`

## Testing
Existing tests

## Review


jorblancoa (approved):

mgeplf:
What is that morphology used for? Is it instantiated in NEURON? I found that with neurolucida ascii files, that a single point wouldn't create a soma, and I imagine that h5 would do the same?

WeinaJi:
Yes, the morphology file is loaded by NEURON in the test. I recall the discussion in https://github.com/neuronsimulator/nrn/issues/2602 that a contour is expected by NEURON. So the new morph .h5 file as a contour should be fine, right? As for  the old h5 file with a single point, we didn't see errors perhaps because the fe exception was ignored by python.

WeinaJi:
I don't know about h5 morph file.  With asc files, we got nan errors from hoc as in [BBPP134-1022](https://bbpteam.epfl.ch/project/issues/browse/BBPP134-1022). As for loading h5 file in python, I would guess the FP errors were ignored in Python and soma was discarded. @pramodk knows more about NEURON.
BTW, with https://github.com/neuronsimulator/nrn/pull/2605, such errors are raised directly.

mgeplf:
I guess it depends on how the h5 is loaded; anyways, just something to be aware of.

PR #105, opened 2023-12-19
Fix memory usage reporting on MacOS

JCGoran:
## Context

MacOS does not use `/proc/`, and consequently Neurodamus crashes when attempting to determine the memory usage of the current process. The fix is to instead use the built-in [`resource`](https://docs.python.org/3/library/resource.html#resource-usage) module to get the memory usage, which works on both Linux and MacOS.

## Review

### DIFF
--- a/neurodamus/utils/memory.py
+++ b/neurodamus/utils/memory.py
@@ -10,7 +10,6 @@ import os
 import json
 import psutil
 import multiprocessing
-from resource import getrusage, RUSAGE_SELF
 
 from ..core import MPI, NeurodamusCore as Nd, run_only_rank0
 
--- a/neurodamus/utils/memory.py
+++ b/neurodamus/utils/memory.py
@@ -10,6 +10,7 @@ import os
 import json
 import psutil
 import multiprocessing
+from resource import getrusage, RUSAGE_SELF
 
 from ..core import MPI, NeurodamusCore as Nd, run_only_rank0
 
@@ -140,9 +141,7 @@ def get_mem_usage_kb():
     """
     Return memory usage information across all ranks, in KiloBytes
     """
-    with open("/proc/self/statm") as fd:
-        _, data_size, _ = fd.read().split(maxsplit=2)
-    usage_kb = float(data_size) * os.sysconf("SC_PAGE_SIZE") / 1024
+    usage_kb = psutil.Process(os.getpid()).memory_info().rss / 1024
 
     return usage_kb
 


alkino (dismissed):
More portability! I love it!

neurodamus/utils/memory.py
```diff
@@ -140,9 +141,7 @@ def get_mem_usage_kb():
     """
     Return memory usage information across all ranks, in KiloBytes
     """
-    with open("/proc/self/statm") as fd:
-        _, data_size, _ = fd.read().split(maxsplit=2)
-    usage_kb = float(data_size) * os.sysconf("SC_PAGE_SIZE") / 1024
+    usage_kb = getrusage(RUSAGE_SELF).ru_maxrss
```
iomaganaris, line 144: It seems to me that this reports the `max RSS` instead of the current one so we have for example in a test of the CI of this PR the following:
```
[INFO] Memusage (RSS) per node [MB]: Max=43942.80, Min=41912.74, Mean(Stdev)=42927.77(1015.03)
[INFO] Memusage (RSS) per task [MB]: Max=1369.22, Min=926.29, Mean(Stdev)=1192.44(106.44)
[INFO] Clearing model
[INFO] malloc_trim: memory released back to the system.
[INFO] Memusage (RSS) per node [MB]: Max=20793.50, Min=20140.71, Mean(Stdev)=20467.10(326.39)
[INFO] Memusage (RSS) per task [MB]: Max=1369.22, Min=926.29, Mean(Stdev)=1192.44(106.44)
```
([source](https://bbpgitlab.epfl.ch/hpc/sim/blueconfigs/-/jobs/1090075))
while in another CI for the same test we have:
```
[INFO] Memusage (RSS) per node [MB]: Max=43950.14, Min=41930.94, Mean(Stdev)=42940.54(1009.60)
[INFO] Memusage (RSS) per task [MB]: Max=1368.47, Min=926.23, Mean(Stdev)=1192.79(106.64)
[INFO] Clearing model
[INFO] malloc_trim: memory released back to the system.
[INFO] Memusage (RSS) per node [MB]: Max=20409.52, Min=19772.78, Mean(Stdev)=20091.15(318.37)
[INFO] Memusage (RSS) per task [MB]: Max=633.22, Min=485.52, Mean(Stdev)=558.09(34.12)
```
([source](https://bbpgitlab.epfl.ch/hpc/sim/blueconfigs/-/jobs/1089505))

iomaganaris (changes requested):

neurodamus/utils/memory.py
JCGoran: You're absolutely right, seems like the two are not equivalent. I opted for a `psutil`-based solution now, it seems to report the same values as reading from `/proc/`, while also working on MacOS.

jorblancoa (approved):

neurodamus/utils/memory.py
iomaganaris: Great! We use `psutil` elsewhere in the code base as well so it should be okay :+1:

iomaganaris (approved):
Welcome change for Mac users :green_apple: :+1:

iomaganaris (approved):

ferdonline (approved):

PR #106, opened 2023-12-19
Fix call to `trim_memory` on non-GNU C libraries

JCGoran:
## Context

The function `malloc_trim` is specific to the GNU C library (glibc), so the call to it in `trim_memory` raises an `AttributeError` when running with other C library implementations (like the default one shipped with MacOS), which we now catch in the `except` block.

## Review


iomaganaris (approved):
LGTM :+1:

jorblancoa (approved):

PR #111, opened 2024-01-12
[BBPBGLIB-1102] Add load balance based on dry run estimate

st4rl3ss:
## Context
This MR includes new mechanisms to generate a detailed load balancing using the memory estimates generated by the dry run. It distributes all the gids of the circuit to a number of ranks specified using the `--num-target-ranks` CLI flag taking care of distributing so that the allocation of each rank is balanced. The system is based on a greedy algorithm, assigning cells in blocks of 10 to the emptiest rank available at that moment.

It also includes automatic pickling and gzipping of the allocation dictionary to `allocation.pkl.gz` file. Loading of the generated load balance will be integrated in a future MR.

## Scope
The MR briefly expands the capabilities of the dry run workflow with automatic load balancing and save to pickle file for future use. All the structure involved stay the same and no significant change is made to workflow.

## Testing
Most of the new functionalities should be automatically tested using the `test_dry_run_workflow` integration tests. In any case a specific test that loads the pickled `allocation.bin` file and checks against expected values was integrated.

## Review

### DIFF
--- a/neurodamus/commands.py
+++ b/neurodamus/commands.py
@@ -55,7 +55,8 @@ def neurodamus(args=None):
         --enable-shm=[ON, OFF]  Enables the use of /dev/shm for coreneuron_input [default: ON]
         --model-stats           Show model stats in CoreNEURON simulations [default: False]
         --dry-run               Dry-run simulation to estimate memory usage [default: False]
-        --prosp-hosts=<number>  Number of prospective hosts for dry-run load balancing [default: 40]
+        --num-target-ranks=<number>  Number of ranks to target for dry-run load balancing
+                                [default: 40]
     """
     options = docopt_sanitize(docopt(neurodamus.__doc__, args))
     config_file = options.pop("ConfigFile")
--- a/neurodamus/commands.py
+++ b/neurodamus/commands.py
@@ -55,6 +55,7 @@ def neurodamus(args=None):
         --enable-shm=[ON, OFF]  Enables the use of /dev/shm for coreneuron_input [default: ON]
         --model-stats           Show model stats in CoreNEURON simulations [default: False]
         --dry-run               Dry-run simulation to estimate memory usage [default: False]
+        --prosp-hosts=<number>  Number of prospective hosts for dry-run load balancing [default: 40]
     """
     options = docopt_sanitize(docopt(neurodamus.__doc__, args))
     config_file = options.pop("ConfigFile")
--- a/neurodamus/node.py
+++ b/neurodamus/node.py
@@ -1868,7 +1868,7 @@ class Neurodamus(Node):
             log_stage("============= DRY RUN (SKIP SIMULATION) =============")
             self._dry_run_stats.display_total()
             self._dry_run_stats.display_node_suggestions()
-            ranks = int(SimConfig.prosp_hosts)
+            ranks = int(SimConfig.num_target_ranks)
             self._dry_run_stats.collect_all_mpi()
             allocation, total_memory_per_rank = distribute_cells(self._dry_run_stats, ranks)
             print_allocation_stats(allocation, total_memory_per_rank)
--- a/neurodamus/node.py
+++ b/neurodamus/node.py
@@ -32,6 +32,7 @@ from .target_manager import TargetSpec, TargetManager
 from .utils import compat
 from .utils.logging import log_stage, log_verbose, log_all
 from .utils.memory import DryRunStats, trim_memory, pool_shrink, free_event_queues, print_mem_usage
+from .utils.memory import print_allocation_stats
 from .utils.timeit import TimerManager, timeit
 from .core.coreneuron_configuration import CoreConfig
 from .io.sonata_config import ConnectionTypes
@@ -1871,16 +1872,7 @@ class Neurodamus(Node):
             ranks = int(SimConfig.prosp_hosts)
             self._dry_run_stats.collect_all_mpi()
             allocation, total_memory_per_rank = distribute_cells(self._dry_run_stats, ranks)
-            # TODO: split this print into a separate function and make it available
-            #       only when logging is on DEBUG level
-            if MPI.rank == 0:
-                print("Allocation: ", allocation)
-                print("Total memory per rank: ", total_memory_per_rank)
-                import statistics
-                values = list(total_memory_per_rank.values())
-                print("Mean: ", statistics.mean(values))
-                print("Median: ", statistics.median(values))
-                print("Stdev: ", statistics.stdev(values))
+            print_allocation_stats(allocation, total_memory_per_rank)
             return
         if not SimConfig.simulate_model:
             self.sim_init()
--- a/neurodamus/node.py
+++ b/neurodamus/node.py
@@ -1865,8 +1865,13 @@ class Neurodamus(Node):
         """
         if SimConfig.dry_run:
             log_stage("============= DRY RUN (SKIP SIMULATION) =============")
+            from .utils.memory import distribute_cells
             self._dry_run_stats.display_total()
             self._dry_run_stats.display_node_suggestions()
+            ranks = 40
+            allocation, total_memory_per_rank = distribute_cells(self._dry_run_stats, ranks)
+            print("Allocation: ", allocation)
+            print("Total memory per rank: ", total_memory_per_rank)
             return
         if not SimConfig.simulate_model:
             self.sim_init()
--- a/neurodamus/node.py
+++ b/neurodamus/node.py
@@ -1868,10 +1868,19 @@ class Neurodamus(Node):
             from .utils.memory import distribute_cells
             self._dry_run_stats.display_total()
             self._dry_run_stats.display_node_suggestions()
-            ranks = 40
+            ranks = int(SimConfig.prosp_hosts)
+            self._dry_run_stats.collect_all_mpi()
             allocation, total_memory_per_rank = distribute_cells(self._dry_run_stats, ranks)
-            print("Allocation: ", allocation)
-            print("Total memory per rank: ", total_memory_per_rank)
+            # TODO: split this print into a separate function and make it available
+            #       only when logging is on DEBUG level
+            if MPI.rank == 0:
+                print("Allocation: ", allocation)
+                print("Total memory per rank: ", total_memory_per_rank)
+                import statistics
+                values = list(total_memory_per_rank.values())
+                print("Mean: ", statistics.mean(values))
+                print("Median: ", statistics.median(values))
+                print("Stdev: ", statistics.stdev(values))
             return
         if not SimConfig.simulate_model:
             self.sim_init()
--- a/neurodamus/node.py
+++ b/neurodamus/node.py
@@ -32,7 +32,7 @@ from .target_manager import TargetSpec, TargetManager
 from .utils import compat
 from .utils.logging import log_stage, log_verbose, log_all
 from .utils.memory import DryRunStats, trim_memory, pool_shrink, free_event_queues, print_mem_usage
-from .utils.memory import print_allocation_stats
+from .utils.memory import print_allocation_stats, export_allocation_stats, distribute_cells
 from .utils.timeit import TimerManager, timeit
 from .core.coreneuron_configuration import CoreConfig
 from .io.sonata_config import ConnectionTypes
@@ -1866,13 +1866,13 @@ class Neurodamus(Node):
         """
         if SimConfig.dry_run:
             log_stage("============= DRY RUN (SKIP SIMULATION) =============")
-            from .utils.memory import distribute_cells
             self._dry_run_stats.display_total()
             self._dry_run_stats.display_node_suggestions()
             ranks = int(SimConfig.prosp_hosts)
             self._dry_run_stats.collect_all_mpi()
             allocation, total_memory_per_rank = distribute_cells(self._dry_run_stats, ranks)
             print_allocation_stats(allocation, total_memory_per_rank)
+            export_allocation_stats(allocation, "allocation.bin")
             return
         if not SimConfig.simulate_model:
             self.sim_init()
--- a/neurodamus/node.py
+++ b/neurodamus/node.py
@@ -32,7 +32,6 @@ from .target_manager import TargetSpec, TargetManager
 from .utils import compat
 from .utils.logging import log_stage, log_verbose, log_all
 from .utils.memory import DryRunStats, trim_memory, pool_shrink, free_event_queues, print_mem_usage
-from .utils.memory import print_allocation_stats, export_allocation_stats, distribute_cells
 from .utils.timeit import TimerManager, timeit
 from .core.coreneuron_configuration import CoreConfig
 from .io.sonata_config import ConnectionTypes
@@ -1870,9 +1869,7 @@ class Neurodamus(Node):
             self._dry_run_stats.display_node_suggestions()
             ranks = int(SimConfig.num_target_ranks)
             self._dry_run_stats.collect_all_mpi()
-            allocation, total_memory_per_rank = distribute_cells(self._dry_run_stats, ranks)
-            print_allocation_stats(allocation, total_memory_per_rank)
-            export_allocation_stats(allocation, "allocation.bin")
+            self._dry_run_stats.distribute_cells(ranks)
             return
         if not SimConfig.simulate_model:
             self.sim_init()
--- a/neurodamus/utils/memory.py
+++ b/neurodamus/utils/memory.py
@@ -12,6 +12,7 @@ import psutil
 import multiprocessing
 import heapq
 import pickle
+import gzip
 
 from ..core import MPI, NeurodamusCore as Nd, run_only_rank0
 from .compat import Vector
@@ -194,8 +195,9 @@ def export_allocation_stats(rank_allocation, filename):
     """
     Export allocation dictionary to serialized pickle file.
     """
+    compressed_data = gzip.compress(pickle.dumps(rank_allocation))
     with open(filename, 'wb') as f:
-        pickle.dump(rank_allocation, f)
+        f.write(compressed_data)
 
 
 @run_only_rank0
@@ -204,7 +206,9 @@ def import_allocation_stats(filename):
     Import allocation dictionary from serialized pickle file.
     """
     with open(filename, 'rb') as f:
-        return pickle.load(f)
+        compressed_data = f.read()
+
+    return pickle.loads(gzip.decompress(compressed_data))
 
 
 @run_only_rank0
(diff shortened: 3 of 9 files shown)


neurodamus/commands.py
```diff
@@ -55,6 +55,7 @@ def neurodamus(args=None):
         --enable-shm=[ON, OFF]  Enables the use of /dev/shm for coreneuron_input [default: ON]
         --model-stats           Show model stats in CoreNEURON simulations [default: False]
         --dry-run               Dry-run simulation to estimate memory usage [default: False]
+        --prosp-hosts=<number>  Number of prospective hosts for dry-run load balancing [default: 40]
```
matz-e, line 58: Can we have a different name for this option? `prosp` seem like a very weird abbreviation. Also, `hosts` would make me think of machine _names_. 

Looking at the rest of the code, this seems to be more the _number of targeted ranks_?

neurodamus/utils/memory.py
```diff
@@ -160,6 +161,159 @@ def pretty_printing_memory_mb(memory_mb):
...
+
+    Returns:
+        rank_allocation (dict): A dictionary where keys are rank IDs and
+                                values are lists of cell IDs assigned to each rank.
+        rank_memory (dict): A dictionary where keys are rank IDs
+                            and values are the total memory load on each rank.
+    """
+    logging.debug("Distributing cells across %d ranks", num_ranks)
+
+    def validate_inputs(dry_run_stats, num_ranks, batch_size):
```
matz-e, line 184: Could this be a method of `DryRunStats`, i.e., we could just call `dry_run_stats.validate()`? Raise a `ValueError` if things go south?

neurodamus/utils/memory.py
```diff
@@ -160,6 +161,159 @@ def pretty_printing_memory_mb(memory_mb):
...
+    import pickle
+    try:
+        with open(filename, 'wb') as f:
+            pickle.dump(rank_allocation, f)
+    except Exception as e:
+        logging.warning("Unable to export allocation stats: {}".format(e))
+
+
+@run_only_rank0
+def import_allocation_stats(filename):
```
matz-e, line 303: Where is this used?

neurodamus/utils/memory.py
```diff
@@ -160,6 +161,159 @@ def pretty_printing_memory_mb(memory_mb):
...
+                yield gid, memory_usage
+
+    # Initialize structures
+    ranks = [(0, i) for i in range(num_ranks)]  # (total_memory, rank_id)
+    heapq.heapify(ranks)
+    rank_allocation = {}
+    rank_memory = {}
+
+    def assign_cells_to_rank(rank_id, batch_memory, rank_allocation, total_memory,
+                             rank_memory, pop):
```
matz-e, line 223:
```suggestion
    def assign_cells_to_rank(rank_id, batch_memory, total_memory, pop):
```
The other variables would be in scope for this function. Simplifies call structure / code a bit.

neurodamus/utils/memory.py
```diff
@@ -160,6 +161,159 @@ def pretty_printing_memory_mb(memory_mb):
...
+        for cell_id, memory in generate_cells(metype_gids):
+            batch.append(cell_id)
+            batch_memory += memory
+            if len(batch) == batch_size:
+                # Get the rank with the lowest memory load
+                total_memory, rank_id = heapq.heappop(ranks)
+                logging.debug("Assigning batch to rank %d", rank_id)
+                # Add the cell to the rank
+                assign_cells_to_rank(rank_id, batch_memory, rank_allocation, total_memory,
+                                     rank_memory, pop)
```
matz-e, line 250: Seems to me that the `heappop` could also be moved inside `assign_cells_to_rank`.

neurodamus/utils/memory.py
```diff
@@ -160,6 +161,159 @@ def pretty_printing_memory_mb(memory_mb):
         return "%.2lf PB" % (memory_mb / 1024 ** 3)
 
 
+@run_only_rank0
+def distribute_cells(dry_run_stats, num_ranks, batch_size=10) -> (dict, dict):
```
matz-e, line 165: Do we have any reasoning for the batch size?

neurodamus/commands.py
st4rl3ss: Yeah, I never liked the name. I used that since it allowed a quick "hijack" of an older unused CLI variable. Will rename to something more akin to what you suggested, thanks!

neurodamus/utils/memory.py
st4rl3ss: The `validate_inputs` function definitely looks ugly there. I'll rework it a bit and integrate it directly into `dry_run_stats` as it seems to make more sense.

neurodamus/utils/memory.py
st4rl3ss: So, at the moment the function is indeed not used anywhere, wrote it as a mirror of the export function for future use when we'll integrate the results into the load balancer. I would delete for the moment, but if you agree, I will at least use it in the workflow test for dryrun.

neurodamus/utils/memory.py
st4rl3ss: Will integrate into next push, thanks!

neurodamus/utils/memory.py
st4rl3ss: The idea of assigning in batches is to (of course) speed up the execution for big circuits. On a 50M cells dataset the assignment took ~5 minutes before and now it's much faster (less than a minute). 

At the moment, 10 is just a number that came out of tests I did on the 50M circuit to obtain a reasonable decrease in execution time without compromising the balancing of the results too much.

As the load balancing function is meant to be used in big circuits mostly, we've decided to compromise the balance in small circuits (that will be much more unbalanced using batch assignment) to obtain a much faster execution for big circuits (where the results will be mostly the same using batch assignment compared to a cell-by-cell assignment).

neurodamus/utils/memory.py
```diff
@@ -160,6 +161,159 @@ def pretty_printing_memory_mb(memory_mb):
...
+        total_memory += batch_memory
+        rank_memory[pop][rank_id] = total_memory
+        # Update total memory and re-add to the heap
+        heapq.heappush(ranks, (total_memory, rank_id))
+
+    # Start distributing cells across ranks
+    for pop, metype_gids in dry_run_stats.metype_gids.items():
+        logging.info("Distributing cells of population %s", pop)
+        rank_allocation[pop] = {}
+        rank_memory[pop] = {}
```
matz-e, line 237: `rank_allocation` and `rank_memory` could be `defaultdict(dict)` to avoid this, too.

st4rl3ss: It helps, thanks. Unfortunately can't remove the:
```
        if rank_id not in rank_allocation[pop]:
            rank_allocation[pop][rank_id] = []
```
later in the code since I would need to declare `rank_allocation` as a `defaultdict(lambda: defaultdict(list))` and this apparently doesn't play well with the generator.

neurodamus/utils/memory.py
```diff
@@ -160,6 +161,159 @@ def pretty_printing_memory_mb(memory_mb):
...
+    # so we just take the biggest value to be safe. (the difference between the two is minimal)
+    average_syns_mem_per_cell = {k: v * 2.0 for k, v in dry_run_stats.average_syns_per_cell.items()}
+
+    # Prepare a list of tuples (cell_id, memory_load)
+    # We sum the memory load of the cell type and the average number of synapses per cell
+    def generate_cells(metype_gids):
+        for cell_type, gids in metype_gids.items():
+            for gid in gids:
+                memory_usage = (dry_run_stats.metype_memory[cell_type] +
+                                average_syns_mem_per_cell[cell_type])
```
matz-e, line 213:
```suggestion

            memory_usage = (dry_run_stats.metype_memory[cell_type] +
                            average_syns_mem_per_cell[cell_type])
            for gid in gids:
```
Computation can be pulled out of the inner loop

st4rl3ss: Uh, that was a remnant of a previous different approach. Thanks for noticing it.

neurodamus/utils/memory.py
```diff
@@ -160,6 +162,136 @@ def pretty_printing_memory_mb(memory_mb):
...
+    Import allocation dictionary from serialized pickle file.
+    """
+    import pickle
+    try:
+        with open(filename, 'rb') as f:
+            rank_allocation = pickle.load(f)
+        return rank_allocation
+    except Exception as e:
+        logging.warning("Unable to import allocation stats: {}".format(e))
+        return None
```
matz-e, line 292: I'm a bit concerned that this will be hiding errors and hard to debug in the future. Can you please

1. Not catch any exceptions
2. Use JSON instead of `pickle` to be able to inspect the stored data easier?

neurodamus/utils/memory.py
```diff
@@ -160,6 +162,136 @@ def pretty_printing_memory_mb(memory_mb):
...
+def export_allocation_stats(rank_allocation, filename):
+    """
+    Export allocation dictionary to serialized pickle file.
+    """
+    import pickle
+    try:
+        with open(filename, 'wb') as f:
+            pickle.dump(rank_allocation, f)
+    except Exception as e:
+        logging.warning("Unable to export allocation stats: {}".format(e))
```
matz-e, line 277: See comment below, same holds here.

neurodamus/utils/memory.py
st4rl3ss: We discussed with @ferdonline about the possibility of using JSON also for this purpose (we are already using it to store the cell memory usage) but then we settled for `pickle` since the allocation table can get very big and we're not expecting the user to view or change the resulting file.

Are there particular usecases where you see the user changing the allocation files? Or any other reason to use JSON instead of `pickle`?

neurodamus/utils/memory.py
ferdonline: This also helps exploiting data locality at synapses reading time.

neurodamus/utils/memory.py
ferdonline: Agree with not handling the Exception here. Maybe we should have some API to check if there is load-balancing data. We may want to have it optional or mandatory controlled by another routine with more context.

matz-e: JSON would make it way easier to write test cases. Also if someone wants to debug something, it'll be easier to handle.

Space and time here are not really that significant, IMO. I ran a small test case with 10 million gids spread over 100 types + 100 ranks. If we're concerned about space, it would be _very_ sensible to gzip everything, since even pickle files are sort of just plain text:
```
❯ ls -lh jayson.json{,.gz} rick.pkl{,.gz}
-rw-r--r-- 1 matwolf bbp  38M Feb  7 15:21 jayson.json
-rw-r--r-- 1 matwolf bbp 249K Feb  7 15:20 jayson.json.gz
-rw-r--r-- 1 matwolf bbp  27M Feb  7 15:22 rick.pkl
-rw-r--r-- 1 matwolf bbp 187K Feb  7 15:06 rick.pkl.gz
```

Using the `ujson` library rather than the built-in `json`, times are also pretty similar:
```
❯ ipython test.ipy
JSON write
507 ms ± 1.22 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)
JSON read
562 ms ± 493 µs per loop (mean ± std. dev. of 7 runs, 1 loop each)
🥒 write
183 ms ± 1.41 ms per loop (mean ± std. dev. of 7 runs, 10 loops each)
🥒 read
427 ms ± 13.6 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)
```
compared to the rest of the setup steps of the simulation, this is all insignificant.

Besides, we don't read the data back in, why write it in the first place? That would reduce the confusing amount of files we scatter everywhere…

ferdonline: Thanks for the measurements, I was expecting numpy pickling to be wiser, maybe something is off?

> Besides, we don't read the data back in

That's the big point here. We will read the data back in, and we want to read it back to numpy arrays and avoid lists. Without pickle can we do that? Honeslty it's nonsense to write complex import function or use third party libs

> JSON would make it way easier to write test cases.

I'm not that convinced. Tests can create the structure easily and deserialization is trivial

neurodamus/utils/memory.py
matz-e: > This also helps exploiting data locality at synapses reading time.

I don't follow. You mean keeping the gids contiguous?

@st4rl3ss I'd be curious if this could be turned into a formula - batch size depending on number of cells + ranks? Maybe later…

neurodamus/utils/memory.py
```diff
@@ -179,11 +234,14 @@ def get_memory_usage(cls, count, synapse_type):
 
 class DryRunStats:
     _MEMORY_USAGE_FILENAME = "cell_memory_usage.json"
+    _ALLOCATION_FILENAME = "allocation.bin"
```
matz-e, line 237:
```suggestion
    _ALLOCATION_FILENAME = "allocation.pkl"
```
At the very minimum. This seems to be a way more canonical ending. Also, can we shove that in a directory together with other simulation byproducts?

neurodamus/utils/memory.py
```diff
@@ -335,3 +395,99 @@ def display_node_suggestions(self):
...
+            # Assign any remaining cells in the last, potentially incomplete batch
+            if batch:
+                assign_cells_to_rank(rank_allocation, rank_memory, batch, batch_memory)
+
+            all_allocation[pop] = rank_allocation
+            all_memory[pop] = rank_memory
+
+        # Print and export allocation stats
+        print_allocation_stats(all_allocation, all_memory)
+        export_allocation_stats(all_allocation, self._ALLOCATION_FILENAME)
```
matz-e, line 478:
```suggestion
            if batch:
                assign_cells_to_rank(rank_allocation, rank_memory, batch, batch_memory)

            all_allocation[pop] = rank_allocation
            all_memory[pop] = rank_memory

        print_allocation_stats(all_allocation, all_memory)
        export_allocation_stats(all_allocation, self._ALLOCATION_FILENAME)
```
These comments don't provide any reasoning, just state what the code already does.

neurodamus/utils/memory.py
```diff
@@ -335,3 +395,99 @@ def display_node_suggestions(self):
...
+            logging.info("Distributing cells of population %s", pop)
+            rank_allocation = defaultdict(Vector)
+            rank_memory = {}
+            batch = []
+            batch_memory = 0
+
+            for cell_id, memory in generate_cells(metype_gids):
+                batch.append(cell_id)
+                batch_memory += memory
+                if len(batch) == batch_size:
```
matz-e, line 464: FWIW, I noticed that in 3.12, `itertools.batched` exists, which would simplify this very nicely.

neurodamus/utils/memory.py
matz-e: > EDIT: Just noticed this is using plain lists. We should try to use a compact data structure

`libsonata.Selection` compresses nicely if gids are continuous. Otherwise, just running gzip on the files seems to compress superbly. And really, if we run on TB sized circuits, 50 MB is nothing.

> We will read the data back in

When? Until that's happening, we should really hold off a bit on dumping more files.

The very minimal expectation should also be that the file format ends in `.pkl` for pickled files. `.bin` could be anything and will be confusing.

st4rl3ss: Following yesterday round of fixes, we've discovered an issue that prevented some of the allocations to be actually pickled. So, even using a more compact data structure (at the moment the `Vector` structure integrated in `neurodamus` which already helped reducing the file size by a factor of 4) the size of the allocation file is ~200 MB for a 50M circuit.

I will experiment with `libsonata.Selection` (as you suggested somewhere else) and `gzip` compression to see if I can improve this.

neurodamus/utils/memory.py
st4rl3ss: Are we ok switching to a 3.12 exclusive functionality?

neurodamus/utils/memory.py
st4rl3ss: I'm pretty sure there is a way to make the batching a bit more dynamic and would be interested in exploring it. 
I will leave it as a further enhancement once we have the whole load balance workflow properly working.

neurodamus/utils/memory.py
st4rl3ss: I've included automatic `gzip` compression so the pickled allocation file for the 50M circuit is down to a more reasonable ~100 MB at basically no performance cost.

neurodamus/utils/memory.py
```diff
@@ -179,11 +238,14 @@ def get_memory_usage(cls, count, synapse_type):
 
 class DryRunStats:
     _MEMORY_USAGE_FILENAME = "cell_memory_usage.json"
+    _ALLOCATION_FILENAME = "allocation.gz"
```
matz-e, line 241:
```suggestion
    _ALLOCATION_FILENAME = "allocation.pkl.gz"
```

neurodamus/utils/memory.py
matz-e: No, this is just FYI

neurodamus/node.py
```diff
@@ -1867,6 +1867,9 @@ def run(self):
             log_stage("============= DRY RUN (SKIP SIMULATION) =============")
             self._dry_run_stats.display_total()
             self._dry_run_stats.display_node_suggestions()
+            ranks = int(SimConfig.num_target_ranks)
```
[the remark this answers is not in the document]
st4rl3ss: `docopt`, that we use to describe the CLI, returns all the arguments as `str` so an explicit conversion is necessary. A validation of the input of the `distribute_cells` functions is still performed in order to catch negative or spurious values.

neurodamus/utils/memory.py
```diff
@@ -179,11 +238,14 @@ def get_memory_usage(cls, count, synapse_type):
...
 class DryRunStats:
     _MEMORY_USAGE_FILENAME = "cell_memory_usage.json"
+    _ALLOCATION_FILENAME = "allocation.pkl.gz"
 
     def __init__(self) -> None:
         self.metype_memory = {}
+        self.average_syns_per_cell = {}
         self.metype_counts = Counter()
         self.synapse_counts = Counter()
+        self.metype_gids = {}
```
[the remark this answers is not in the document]
st4rl3ss: Not sure what you mean. 😅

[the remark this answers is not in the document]
ferdonline: gids per metype. I think we keep the gids sorted

ferdonline (approved):

neurodamus/utils/memory.py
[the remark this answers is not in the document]
matz-e: sort the lines :)

ferdonline (approved):

PR #113, opened 2024-01-22
[BBPBGLIB-1114] Apply offset to the gids before reading node file

jorblancoa:
## Context
Fixes longrun ci issue detected [here](https://bbpteam.epfl.ch/project/issues/browse/BBPBGLIB-1114).

## Scope
Apply population offset to the gids before reading node file when reusing load balancer.

## Review

### DIFF
--- a/neurodamus/cell_distributor.py
+++ b/neurodamus/cell_distributor.py
@@ -249,7 +249,9 @@ class CellManagerBase(_CellManager):
         logging.info(" -> Distributing target '%s' using Load-Balance", target_spec.name)
         self._binfo = load_balancer.load_balance_info(target_spec)
         # self._binfo has gidlist, but gids can appear multiple times
-        all_gids = numpy.unique(self._binfo.gids.as_numpy().astype("uint32"))
+        all_gids = numpy.unique(
+            self._binfo.gids.as_numpy().astype("uint32") - self._local_nodes.offset
+        )
         total_cells = len(all_gids)
         gidvec, me_infos, full_size = loader_f(self._circuit_conf, all_gids)
         return gidvec, me_infos, total_cells, full_size


neurodamus/cell_distributor.py
```diff
@@ -249,7 +249,9 @@ def _load_nodes_balance(self, loader_f, load_balancer):
         logging.info(" -> Distributing target '%s' using Load-Balance", target_spec.name)
         self._binfo = load_balancer.load_balance_info(target_spec)
         # self._binfo has gidlist, but gids can appear multiple times
-        all_gids = numpy.unique(self._binfo.gids.as_numpy().astype("uint32"))
+        all_gids = numpy.unique(
```
WeinaJi, line 252: If I understand correctly, the loadbalance dat file stored in `sim_conf` are based on gids+offset and here it removes it. Is it better the other way around, that we store based gids in the ld data? Does it require lots of changes? @ferdonline

jorblancoa: Isn't that what @ferdonline mentioned during the weekly sync? that he prefers to store the ids without taking into account the offset?

jorblancoa: I guess it would be here?
```
        target_gids = self._get_target_gids(target_spec)
        .....
        # Write the new cx file since Neuron needs it to do CPU assignment
        with open(new_cx_filename, "w") as newfile:
            self._write_msdat_dict(newfile, cx_other, target_gids - **offset**)
```
What do you guys prefer?

jorblancoa: actually this creates more issues later on...so i think it is better to leave it with my initial fix

ferdonline: I still have to investigate the code better. If I understand you right, Weina, you mention that the `_binfo.gids` should not require offsets. My main concern is the files written to disk, so they can be combined. If this `_binfo` is a direct representation of the files on disk then I think we have something to fix.

Anyway, since Jorge mentioned, this might be something non-trivial. We can merge the quick fix and create a ticket to improve that properly. However I don't think the will have a high priority, since we dont reuse load balance across simulations yet

WeinaJi: Right, my motivation to storing the raw gids in the loadbalancing dat files are reusability in other circuits, but it is not a high priority as @ferdonline said. There would require more changes I am afraid.
 
Let's merge this fix.

WeinaJi (approved):

PR #114, opened 2024-01-24
[BBPBGLIB-1115] Fix issue with summation reports and cell target in NEURON

jorblancoa:
## Context
When using NEURON as simulator and having a summation report with a cell target or "sections": "soma", certain sections, such as myelin, are not included in the summation.

## Scope
Remove hoc getPointList() call, use python version instead.

## Review


neurodamus/target_manager.py
```diff
@@ -225,7 +225,7 @@ def init_hoc_manager(self, cell_manager):
         # give a TargetManager the TargetParser's completed targetList
         self.hoc = Nd.TargetManager(self.parser.targetList, cell_manager)
 
-    def get_target_points(self, target, cell_manager, cell_use_compartment_cast, **kw):
+    def get_target_points(self, target, cell_manager, **kw):
```
WeinaJi, line 228: Is `cell_use_compartment_case` not needed any more? If so, we can remove it from the docstring and in `node.py`

WeinaJi (approved):

PR #115, opened 2024-01-29
Atemerev/opts config

atemerev:
## Context
Options handling by `init.py`: `special` only supports positional options. Attempting to specify a config file as a positional option led to confusing error messages.

Now, using positional options in init.py lead to an error message explaining that only keyword options are supported. --configFile can be used anywhere on the command line, the rest of the options are passed correctly to `neurodamus.commands`.

## Scope
Init.py is affected, no new dependencies added. Handling arguments is extracted to a function for unit testing.

## Testing
tests/unit/test_init_py_config.py

## Review
* [X] Unit/Scientific test added

### DIFF
--- a/init.py
+++ b/init.py
@@ -9,24 +9,30 @@ from neurodamus import commands
 from neuron import h
 
 
-def main():
+def extract_arguments(args):
     """Get the options for neurodamus and launch it.
 
     We can't use positional arguments with special so we look for
     --configFile=FILE, which defaults to simulation_config.json
     """
     config_file = "simulation_config.json"
-
-    for i, arg in enumerate(sys.argv[1:]):
+    for i, arg in enumerate(args[1:]):
         if not arg.startswith("-"):
-            print("Positional arguments are not supported by init.py. Please specify --configFile=<config file> to\n"
-                  "run this script (or leave empty to use the default, ./simulation_config.json).", file=sys.stderr)
-            sys.exit(1)
+            raise ValueError("Positional arguments are not supported")
         elif arg.startswith("--configFile="):
             config_file = arg.split('=')[1]
             break
+    args = [config_file] + [x for x in args[1:] if not x.startswith("--configFile=")]
+    return args
 
-    args = [config_file] + [x for x in sys.argv[1:] if not x.startswith("--configFile=")]
+
+def main():
+    try:
+        args = extract_arguments(sys.argv)
+    except ValueError:
+        print("Positional arguments are not supported by init.py. Please specify --configFile=<config file> to\n"
+              "run this script (or leave empty to use the default, ./simulation_config.json).", file=sys.stderr)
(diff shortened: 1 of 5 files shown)


pyproject.toml
```diff
@@ -43,6 +43,9 @@ exclude=["tests"]
 [tool.pytest.ini_options]
 addopts = "--verbose"
 markers = ["slow: marks tests as slow"]
+pythonpath = [
```
ferdonline, line 46: I would prefer to not change a default affecting the whole pytest execution. Maybe you could use importlib to have direct access to the "external" module/function you want to test

atemerev: That's only for unit tests, but OK, I will see what could be done about it.

init.py
```diff
@@ -6,27 +6,32 @@
...
+    args = [config_file] + [x for x in args[1:] if not x.startswith("--configFile=")]
+    return args
 
-    args = [config_file] + sys.argv[first_argument_pos:]
+
+def main():
+    try:
+        args = extract_arguments(sys.argv)
+    except ValueError:
+        print("Positional arguments are not supported by init.py. Please specify --configFile=<config file> to\n"
```
WeinaJi, line 32: I would prefer to let neuron `h.quit(exitcode)` to handle the proper abort in particular for MPI processes. How about `logging.error` and return a code here ?

atemerev: done

WeinaJi: > The only thing, Im not sure logging is configured at this point

At this point, logging just simply throws `ERROR:root:Positional arguments are not supported ....`
We also have an early call in `command.py` . 
```
    if not os.path.isfile(config_file):
        logging.error("Config file not found: %s", config_file)
        return 1
```
That seems ok, right? @ferdonline

tests/unit/conftest.py
```diff
@@ -1,6 +1,7 @@
 import pytest
 import sys
 import unittest
+import unittest.mock
```
WeinaJi, line 4: Is it sufficient to just import either of these two lines?

neurodamus/utils/cli.py
WeinaJi, line 1: Thank you for the implementation. It looks good.
One thing, can we block the use case `special -mpi -python --configFile=simulation_sonata.json $NEURODAMUS_PYTHON/init.py`? This is the current behaviour of the main branch. What do you think?

atemerev: So, specifying the config file before init.py should raise an error?

WeinaJi: yes, I would think so. Maybe we could modify the error msg in `init.py` to be something like "... please specify init.py --configFile=<config> ..." ?

atemerev: Okay!

WeinaJi (approved):

PR #118, opened 2024-02-02
[BBPBGLIB-1124] Remove hoc targets related code

jorblancoa:
## Context
Addresses BBPBGLIB-1124.
Removes code related to hoc targets, as well as the need of using a hoc TargetManager.

## Scope
Few functions were implemented in the python Targetmanager in order to remove the hoc version:
getMETypes(), getPointList() and location_to_point()

## Review


neurodamus/target_manager.py
```diff
@@ -269,9 +220,100 @@ def pathways_overlap(self, conn1, conn2, equal_only=False):
...
+            if section.orientation() == 1:
+                ipt = section.n3d() - 1 - ipt
+                offset = -offset
+
+            if ipt < section.n3d():
+                distance = (section.arc3d(int(ipt)) + offset) / section.L
+                if distance == 0.0:
+                    distance = 0.0000001
+                if distance >= 1.0:
+                    distance = 0.9999999
```
WeinaJi, line 310: I am wondering if we can use the `max(min(...)...)` as you did above for `offset`. But is it likely that distance < 0 after calculation?

neurodamus/target_manager.py
```diff
@@ -269,9 +220,100 @@ def pathways_overlap(self, conn1, conn2, equal_only=False):
...
+            raise Exception(f"Error: section {isec} out of bounds ({cell_sections.num_sections} "
+                            "total). Morphology section count is low, is this a good morphology?")
+
+        distance = 0.5
+        tmp_section = cell_sections.isec2sec[int(isec)]
+
+        if tmp_section is None:  # Assume we are in LoadBalance mode
+            result_point.append(None, -1)
+        elif ipt == -1:
+            # SYN2/Sonata spec have a pre-calculated distance field.
```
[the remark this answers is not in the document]
WeinaJi: And looking at `synaspe_reader.py`, ipt = -1 , offset =  afferent_section_pos = [0, 1] according to the SPEC https://sonata-extension.readthedocs.io/en/latest/sonata_tech.html#fields-for-edges, shall we just keep the SONATA case here?

neurodamus/target_manager.py
```diff
@@ -600,118 +599,31 @@ def generate_subtargets(self, n_parts):
         return [[targets[cycle_i] for targets in new_targets.values()]
                     for cycle_i in range(n_parts)]
 
+    def set_offset(self, *_):
+        pass  # Only hoc targets require manually setting offsets
```
WeinaJi, line 603: Do we still need to `set_offset`?

neurodamus/target_manager.py
```diff
@@ -269,9 +220,100 @@ def pathways_overlap(self, conn1, conn2, equal_only=False):
             return TargetSpec(src1) == TargetSpec(src2) and TargetSpec(dst1) == TargetSpec(dst2)
         return self.intersecting(src1, src2) and self.intersecting(dst1, dst2)
 
-    def __getattr__(self, item):
-        logging.debug("Compat interface to TargetManager::" + item)
-        return getattr(self.hoc, item)
+    def getPointList(self, target, **kw):
+        return self.get_target_points(self._targets[target], self._cell_manager, **kw)
```
WeinaJi, line 224: Now we call the chain
```
TargetManager.getPointList(target, cell_manager) =>
     TargetManager.get_target_points(target, cell_manager =>
          target.getPointList(cell_manager)
```
Shall we remove the middle call `TargetManager.get_target_points` ?

WeinaJi (commented):
Thank you for the great work! I have some comments, and I am open for discussions.

neurodamus/target_manager.py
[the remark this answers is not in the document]
jorblancoa: you mean to keep only the ipt==-1 case? and remove the else?
Or rename the offset to afferent_section_pos?

neurodamus/target_manager.py
jorblancoa: you are right, this was called from few places but I guess was only for hoc targets, will remove that also

neurodamus/target_manager.py
jorblancoa: makes sense, but i will then rename the first call to get_target_points() and only have:
TargetManager.get_target_points(target, cell_manager) =>
          target.getPointList(cell_manager)

examples/test_neurodamus.py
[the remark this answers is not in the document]
jorblancoa: Maybe we should add a file in the examples directory? or point to an existing simulation under tests like usecase3?

WeinaJi (approved):

PR #120, opened 2024-02-05
[BBPBGLIB-1110] Error now raised when executeConfigure is returned with errors

atemerev:
## Context

In the simulation config, synapse_configure contains HOC code. Neurodamus should fail fast when there is an error in configuration, instead of passing it forward.

## Scope

Error check is added to connection.py

## Testing

TBD

## Review


tests/integration-e2e/test_connection.py
```diff
@@ -6,6 +6,7 @@
 
 SIM_DIR = Path(__file__).parent.parent.absolute() / "simulations" / "v5_sonata"
 CONFIG_FILE_MINI = SIM_DIR / "simulation_config_mini.json"
+CONFIG_FILE_ERR = SIM_DIR / "simulation_config_err.json"
```
WeinaJi, line 9: Can we create a temp config file using the basic data `sonata_config()` from `conftest.py` ? Maybe we can set fixtures like in `test_mulitpop.py` ? 
```
@pytest.fixture
def sonata_config_file(sonata_config, extra_config):
```

neurodamus/connection.py
```diff
@@ -557,12 +557,17 @@ def _update_conductance_ratio(self, syn_obj, is_inhibitory, value):
         else:
             syn_obj.NMDA_ratio = value
 
+    def _configure_hoc(self, synapses, configuration):
```
jorblancoa, line 560: I would prefer another name for this method.
Maybe _configure_synapses()?

jorblancoa:
Thanks for the PR!
Maybe this could be tackled in another PR, but since we are cleaning .hoc files whenever possible, maybe the ConnectionUtils.hoc could be removed and the executeConfigure() function ported to python.
At first glance it doesn't seem to be complicated. The only problematic call would be execute1(), and we could use Nd.execute1() from python.

What do you think? @WeinaJi

WeinaJi:
> Thanks for the PR! Maybe this could be tackled in another PR, but since we are cleaning .hoc files whenever possible, maybe the ConnectionUtils.hoc could be removed and the executeConfigure() function ported to python. At first glance it doesn't seem to be complicated. The only problematic call would be execute1(), and we could use Nd.execute1() from python.
> 
> What do you think? @WeinaJi

Yes, sounds good. There is already an example in https://github.com/BlueBrain/neurodamus/blob/main/neurodamus/node.py#L1152
Shall we add this python function directly in `connection.py`?

jorblancoa:
> > Thanks for the PR! Maybe this could be tackled in another PR, but since we are cleaning .hoc files whenever possible, maybe the ConnectionUtils.hoc could be removed and the executeConfigure() function ported to python. At first glance it doesn't seem to be complicated. The only problematic call would be execute1(), and we could use Nd.execute1() from python.
> > What do you think? @WeinaJi
> 
> Yes, sounds good. There is already an example in https://github.com/BlueBrain/neurodamus/blob/main/neurodamus/node.py#L1152 Shall we add this python function directly in `connection.py`?

I think it makes sense yes

atemerev:
> > > Thanks for the PR! Maybe this could be tackled in another PR, but since we are cleaning .hoc files whenever possible, maybe the ConnectionUtils.hoc could be removed and the executeConfigure() function ported to python. At first glance it doesn't seem to be complicated. The only problematic call would be execute1(), and we could use Nd.execute1() from python.
> > > What do you think? @WeinaJi
> > 
> > 
> > Yes, sounds good. There is already an example in https://github.com/BlueBrain/neurodamus/blob/main/neurodamus/node.py#L1152 Shall we add this python function directly in `connection.py`?
> 
> I think it makes sense yes

OK, so, what's the resolution? I'll delete this PR and try to work on the python port instead?

WeinaJi:
How about continuing in the same PR to convert `ConnectionUtils.hoc` to Python function? You can modify the PR title and description.

jorblancoa:
After speaking with @ferdonline, he mentioned that porting this to python would be harder than I anticipated, so we can leave it for another time!
@atemerev Forget my last comment! I think we can merge this

tests/integration-e2e/test_connection.py
atemerev: done

atemerev:
@jorblancoa OK, let's merge then!

jorblancoa (approved):

PR #121, opened 2024-02-06
[BBPBGLIB-1125] GH forkflow for building a wheel

atemerev:
## Context

Build and upload a PyPI wheel

## Scope

New Github actions workflow

## Testing

Should see if it works with Github actions

## Review


.github/workflows/publish_wheels.yml
```diff
@@ -0,0 +1,58 @@
...
+          cd ..
+      - name: Install neuron
+        run: |
+          pip3 install neuron
+      - name: Build Wheel
+        run: |
+          cd $GITHUB_WORKSPACE
+          python${{ matrix.python_version }} -m pip install --upgrade build
+          # build wheel
+          python${{ matrix.python_version }} -m build .
```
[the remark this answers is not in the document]
matz-e: Yes, from a Spack perspective, a source tarball is very desirable. Also for any system not covered by manylinux.

.github/workflows/publish_wheels.yml
```diff
@@ -0,0 +1,46 @@
...
+          fetch-depth: 500
+          fetch-tags: true
+      - name: Prerequisites
+        run: |
+          sudo apt-get update --fix-missing
+          sudo apt-get install -y software-properties-common
+          sudo add-apt-repository -y ppa:deadsnakes/ppa
+          sudo TZ=UTC apt-get -y install tzdata
+          sudo apt-get update 
+          sudo apt-get install -y python${{ matrix.python_version }} python${{ matrix.python_version }}-distutils python${{ matrix.python_version }}-venv
```
WeinaJi, line 24: Do we still use distutils?

atemerev: No, no distutils now.

.github/workflows/publish_wheels.yml
```diff
@@ -0,0 +1,46 @@
...
+          cd $GITHUB_WORKSPACE
+          python${{ matrix.python_version }} -m pip install --upgrade build
+          # build wheel
+          python${{ matrix.python_version }} -m build .
+      - name: Test Wheel
+        run: |
+          cd $GITHUB_WORKSPACE
+          pip3 install tox
+          DIST_FILE=`ls dist/*whl` && tox --installpkg=$DIST_FILE -e ${{ matrix.python_version }}
+      - name: Upload Wheel
```
WeinaJi, line 39: I think for uploading wheel, we need the rule like if pushing to tags

WeinaJi (commented):
Please also add a

.github/workflows/publish_wheels.yml
```diff
@@ -0,0 +1,43 @@
+name: Neurodamus Wheels
+
+on:
+  push:
+    tags:
+      - '*'
+  workflow_dispatch:
```
WeinaJi, line 7: Shall we set different rules for build, test and upload?
We build and test the distribution files when having pull requests to `main`
We build/test/upload when pushing to `tags` 

In this way, each change in the PR is tested for building the wheel. 

That's what we do currently in the GitLab CI although we upload to the internal `devpi`.

WeinaJi: Sorry, I may have not described my request well. I would prefer to build, test and push in the same workflow, different steps. If we do the pushing in a different workflow, we would have to check whether the wheel is built before pushing.

I think you can set the rule to activate the push `step` using `if: github.event_name == 'push' && startsWith(github.ref, 'refs/tags')`like in the [libsonata workflow](https://github.com/BlueBrain/libsonata/blob/0f7b4e4232c76d118de64dfbfe948ecc7d1c7780/.github/workflows/publish-sdist-wheels.yml) .

WeinaJi:
looks good to me. @jorblancoa , can you also have a look ?

jorblancoa:
looks good to me also!

jorblancoa (approved):

WeinaJi (approved):

PR #124, opened 2024-02-14
Add tests for python 3.9 to 3.12

jorblancoa:
## Context
Improve github actions CI by adding python versions 3.9 to 3.12.

## Review

### DIFF
--- a/.github/workflows/simulation_test.yml
+++ b/.github/workflows/simulation_test.yml
@@ -27,14 +27,31 @@ env:
 
 jobs:
   simulation:
-    runs-on: ubuntu-22.04
+    runs-on: ${{ matrix.os }}
 
+    strategy:
+      matrix:
+        os: [ubuntu-22.04]
+        python-version: [3.8, 3.9, 3.10, 3.11, 3.12]
     steps:
+    - name: Setup Python
+      uses: actions/setup-python@v5
+      with:
+        python-version: ${{ matrix.python-version }}
+
     - name: Checkout neurodamus repository
       uses: actions/checkout@v4
       with:
         ref: ${{ github.event.pull_request.head.sha }}
         fetch-depth: 0
+
+    - name: Cache pip
+      uses: actions/cache@v3
+      with:
+        path: ~/.cache/pip
+        key: ${{ matrix.os }}-pip-${{ matrix.python-version }}-v1
+        restore-keys: |
+          ${{ matrix.os }}-pip-
 
     - name: Get HEAD commit message and look for branches
       run: |
@@ -45,10 +62,15 @@ jobs:
         LIBSONATA_REPORT_BRANCH=$(echo $COMMIT_MESSAGE | grep -Po 'LIBSONATA_REPORT_BRANCH=\K[0-9a-zA-Z/_.\-]*' || true)
         if [[ ! -z $LIBSONATA_REPORT_BRANCH ]]; then echo "LIBSONATA_REPORT_BRANCH=$LIBSONATA_REPORT_BRANCH" >> $GITHUB_ENV; fi
 
-    - name: Install hdf5-mpich
+    - name: Install system dependencies
       run: |
         sudo apt-get update
-        sudo apt-get install mpich libmpich-dev libhdf5-mpich-dev hdf5-tools
+        sudo apt-get install -y mpich libmpich-dev libhdf5-mpich-dev hdf5-tools flex libfl-dev bison ninja-build
+
+    - name: Upgrade pip and install base Python packages
+      run: |
+        python -m pip install --upgrade pip setuptools
+        pip install cython numpy wheel pkgconfig
 
     - name: Cache python virtual env
       id: cache-venv
@@ -100,17 +122,25 @@ jobs:
         export SONATAREPORT_DIR=$(pwd)/libsonatareport/build/install
         sudo apt-get install flex libfl-dev bison ninja-build
         . ./venv/bin/activate
-        git clone --branch=${{ env.NEURON_BRANCH }} https://github.com/neuronsimulator/nrn.git
+        git clone --branch=${{ env.NEURON_BRANCH }} https://github.com/neuronsimulator/nrn.git --recursive --shallow-submodules --depth=1
         cd nrn
         if [[ ${{ env.NEURON_BRANCH }} == 'master' ]]; then
           git checkout ${{ env.NEURON_COMMIT_ID }}
         fi
-        python -m pip install --upgrade pip -r nrn_requirements.txt
-        mkdir build && cd build
-        cmake -G Ninja -DPYTHON_EXECUTABLE=$(which python) -DCMAKE_INSTALL_PREFIX=$(pwd)/install -DNRN_ENABLE_MPI=ON -DNRN_ENABLE_INTERVIEWS=OFF \
-          -DNRN_ENABLE_CORENEURON=ON -DCMAKE_C_COMPILER=gcc -DCMAKE_CXX_COMPILER=g++ -DCORENRN_ENABLE_REPORTING=ON -DCMAKE_PREFIX_PATH=$SONATAREPORT_DIR ..
-        cmake --build . --parallel
-        cmake --build . --target install
+        cd ..
+        python -m pip install --upgrade pip -r nrn/nrn_requirements.txt
+        cmake -B nrn/build -S nrn -G Ninja \
+          -DPYTHON_EXECUTABLE=$(which python) \
+          -DCMAKE_INSTALL_PREFIX=$(pwd)/nrn/build/install \
+          -DNRN_ENABLE_MPI=ON \
+          -DNRN_ENABLE_INTERVIEWS=OFF \
+          -DNRN_ENABLE_CORENEURON=ON \
+          -DCMAKE_C_COMPILER=gcc \
+          -DCMAKE_CXX_COMPILER=g++ \
+          -DCORENRN_ENABLE_REPORTING=ON \
+          -DCMAKE_PREFIX_PATH=$SONATAREPORT_DIR
+        cmake --build nrn/build --parallel
+        cmake --build nrn/build --target install
 
     - name: Install mpi4py
       if: steps.cache-venv.outputs.cache-hit != 'true'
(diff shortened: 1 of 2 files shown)


.github/workflows/simulation_test.yml
```diff
@@ -55,22 +77,27 @@ jobs:
...
+          -DCMAKE_INSTALL_PREFIX=$(pwd)/install \
+          -DNRN_ENABLE_MPI=ON \
+          -DNRN_ENABLE_INTERVIEWS=OFF \
+          -DNRN_ENABLE_CORENEURON=ON \
+          -DCMAKE_C_COMPILER=gcc \
+          -DCMAKE_CXX_COMPILER=g++ \
+          -DCORENRN_ENABLE_REPORTING=ON \
+          -DCMAKE_PREFIX_PATH=$SONATAREPORT_DIR ..
         cmake --build . --parallel
         cmake --build . --target install
```
matz-e, line 97:
```suggestion
        cmake -B nrn/build -S nrn -G Ninja \
          -DPYTHON_EXECUTABLE=$(which python) \
          -DCMAKE_INSTALL_PREFIX=$(pwd)/nrn/build/install \
          -DNRN_ENABLE_MPI=ON \
          -DNRN_ENABLE_INTERVIEWS=OFF \
          -DNRN_ENABLE_CORENEURON=ON \
          -DCMAKE_C_COMPILER=gcc \
          -DCMAKE_CXX_COMPILER=g++ \
          -DCORENRN_ENABLE_REPORTING=ON \
          -DCMAKE_PREFIX_PATH=$SONATAREPORT_DIR
        cmake --build nrn/build --parallel
        cmake --build nrn/build --target install
```

ferdonline:
Contrary to what was expected it seems that it is not due to the destructor.
I start to think that it's something specific to this system, namely the MPI4Py version.
@WeinaJi I see this is building Mpi4py from source. Do you think there's a way we can install a prebuilt one? It may save us from this problem

WeinaJi:
@ferdonline , it seems that we can use `pip install mpi4py` while rebuilding `h5py`. I tested with the docker image. Can you change that in the GH actions script?

WeinaJi:
I've push a change to use prebuilt `mpi4py` but the error is still there.

neurodamus/node.py
```diff
@@ -1536,30 +1533,29 @@ def cleanup(self):
...
+    @run_only_rank0
+    def _delete_corenrn_data(self):
+        data_folder = SimConfig.coreneuron_datadir
+        logging.info("Deleting intermediate data in %s", data_folder)
+
+        if ospath.islink(data_folder):
+            # in restore, coreneuron data is a symbolic link
+            os.unlink(data_folder)
+        else:
+            subprocess.call(['/bin/rm', '-rf', data_folder])
```
WeinaJi, line 1549: Is the fork process created here? From the failed job log, it seems we either set `RDMAV_FORK_SAFE` or delete the foler without fork(). For this test purpose, it seems `RDMAV_FORK_SAFE` is sufficient. 
```
For the libfabric EFA provider to work safely when fork()
is called, you will need to set the following environment
variable:
          RDMAV_FORK_SAFE

However, setting this environment variable can result in
signficant performance impact to your application due to
increased cost of memory registration.

You may want to check with your application vendor to see
if an application-level alternative (of not using fork)
exists.
```

matz-e (approved):

neurodamus/node.py
matz-e: Why use `subprocess.call` and not something like https://docs.python.org/3/library/shutil.html#shutil.rmtree ?

WeinaJi: I recalled that we used `shutil.rmtree` at the beginning but it was too slow to delete a folder with a large amount of files https://bbpteam.epfl.ch/project/issues/browse/BBPBGLIB-540 . Then we changed to use "rm -rf" from `subprocess.call`.

ferdonline (dismissed):
Looks good to me

@jorblancoa @WeinaJi How do you feel about the changes I did trying to simplify the destructor? We now know it wasn't the problem in this case. Shall we keep it anyway?

.github/workflows/simulation_test.yml
```diff
@@ -55,23 +77,27 @@ jobs:
...
+          -DCMAKE_INSTALL_PREFIX=$(pwd)/nrn/build/install \
+          -DNRN_ENABLE_MPI=ON \
+          -DNRN_ENABLE_INTERVIEWS=OFF \
+          -DNRN_ENABLE_CORENEURON=ON \
+          -DCMAKE_C_COMPILER=gcc \
+          -DCMAKE_CXX_COMPILER=g++ \
+          -DCORENRN_ENABLE_REPORTING=ON \
+          -DCMAKE_PREFIX_PATH=$SONATAREPORT_DIR
+        cmake --build nrn/build --parallel
+        cmake --build nrn/build --target install
```
matz-e, line 96: Can we also cache the `nrn` directory? Maybe pin a commit as the key?

It seems that neurodamus should not be _that_ reliant on NRN master. Caching `libsonata`, `libsonatareport` and `nrn` builds could save 7-10 minutes build time and conserve runner minutes.

jorblancoa:
> Looks good to me
> 
> @jorblancoa @WeinaJi How do you feel about the changes I did trying to simplify the destructor? We now know it wasn't the problem in this case. Shall we keep it anyway?

For me the changes seem reasonable!

.github/workflows/simulation_test.yml
WeinaJi: Yes, caching neuron will speed up a lot. It will be nice to cache the same commit as the neuron spack recipe but we can't know the ID before launching the workflow, right? Then we need to hardcode a version and update it manually?

neurodamus/node.py
```diff
@@ -1592,12 +1588,8 @@ def __init__(
                 1 - Info messages (default)
                 2 - Verbose
                 3 - Debug messages
-            cleanup_atexit: (bool) Call cleanup in the destructor
-                [for more see: https://bbpteam.epfl.ch/project/issues/browse/BBPBGLIB-976]
```
WeinaJi, line 1596: I think better to keep `cleanup_atexit` because I am using it for Polina's gap junction compensation
```
 nd = Neurodamus(config_file, auto_init=False, logging_level=log_level, cleanup_atexit=False, **options)
__ = load_user_modification(nd, node_pop="thalamus_neurons") # gj compensation
nd.init()
nd.run()
nd.cleanup()
```
I see that multiscale_run calls `cleanup_atexit=false` as well, https://bbpgitlab.epfl.ch/molsys/multiscale_run/-/blob/main/multiscale_run/neurodamus_manager.py?ref_type=heads#L20 . We don't know why we need to clean up later, as address in BBPBGLIB-976.

ferdonline: `cleanup_atexit=false` is just a workaround introduced to avoid calling the destructor because it was creating problems. My change fixes exactly that case. We shouldn't ever avoid calling the destructor, the destructor should be safe. So now what we do is to call the cleanup procedure outside.

WeinaJi: ok, I understand better now.  Thank for the improvement. We can merge it together in this PR.  

BTW, then probably we can remove `cleanup_atexit=True` from the `Neurodamus` class constructor ? Also ping multiscale_run ppl to change their script.

ferdonline: Oh, yes, it was an oversight to leave it there

.github/workflows/simulation_test.yml
WeinaJi: Sure. I will add the caching in a separate PR https://github.com/BlueBrain/neurodamus/pull/166 .

.github/workflows/simulation_test.yml
```diff
@@ -55,23 +77,27 @@ jobs:
...
         export SONATAREPORT_DIR=$(pwd)/libsonatareport/build/install
+        git clone https://github.com/BlueBrain/nmodl.git --depth=1
         git clone --branch="${NEURON_BRANCH:-master}" https://github.com/neuronsimulator/nrn.git --depth=1
-        cd nrn
-        mkdir build && cd build
-        cmake -G Ninja -DPYTHON_EXECUTABLE=$(which python) -DCMAKE_INSTALL_PREFIX=$(pwd)/install -DNRN_ENABLE_MPI=ON -DNRN_ENABLE_INTERVIEWS=OFF \
-          -DNRN_ENABLE_CORENEURON=ON -DCMAKE_C_COMPILER=gcc -DCMAKE_CXX_COMPILER=g++ -DCORENRN_ENABLE_REPORTING=ON -DCMAKE_PREFIX_PATH=$SONATAREPORT_DIR ..
-        cmake --build . --parallel
-        cmake --build . --target install
+        python -m pip install --upgrade pip -r nrn/nrn_requirements.txt
```
WeinaJi, line 83: Is it needed to install the nmodl dependencies ? We don't do that when building the docker container and it seems working.

WeinaJi:
> I'm getting the impression that there are two threads in this PR, one for the CI improvements and another one around the destructor and we would probably do better to separate both. I think it became more apparent as we are still tuning the CI. If you don't have anything against I can separate them

Right, it will look better in the git history to merge them separately. The destructor part looks good to me. If you separate it in another PR I will approve it. As for tuning CI, I will add the caching in #166 . This PR could remain as tests for multiple python versions.

.github/workflows/simulation_test.yml
```diff
@@ -17,15 +17,32 @@ on:
...
       uses: actions/checkout@v4
       with:
         ref: ${{ github.event.pull_request.head.sha }}
         fetch-depth: 0
 
+    - name: Cache pip
+      uses: actions/cache@v3
+      with:
+        path: ~/.cache/pip
+        key: ${{ runner.os }}-pip-${{ matrix.python-version }}-v1
```
WeinaJi, line 42: Here `${{ runner.os }}` = `Linux` , but I feel it is not sufficient for the key. How about using ${{ matrix.os }} which is `ubuntu-22.04` ?

WeinaJi (commented):
More caching will come from https://github.com/BlueBrain/neurodamus/pull/166. This PR can be merged first.

.github/workflows/simulation_test.yml
jorblancoa: sounds good!

WeinaJi:
@matz-e @ferdonline , caching and multiple python versions are in place. Shall we merge it?

WeinaJi (approved):

PR #129, opened 2024-02-21
[BBPGLIB-1102] (Part two) Load memory balance from file

st4rl3ss:
## Context
Addresses the second part of https://bbpteam.epfl.ch/project/issues/browse/BBPBGLIB-1102 by loading and using the allocation data generated via dry run to memory balance execution.

## Scope
No present structures are changed, only added.
A new load balancing mode is introduced by using the `--lb-mode=Memory` flag in order to load and use the `allocation.pkl.gz` file generated in a previous run.
Also now the dry run workflow automatically generates an `allocation.pkl.gz` file with the number of ranks equal to the number of suggested nodes multiplied by the amount of physical core present in the current machine.

## Testing
A new test specific for the Load Balance Memory workflow was added.
Also the `dry-run` workflow test was expanded with a 1-rank distribution, in order to also generate an `allocation.pkl.gz` file that can be ran for the Load Balance Memory workflow.

## Review


neurodamus/cell_distributor.py
```diff
@@ -201,11 +201,13 @@ def load_nodes(self, load_balancer=None, *, _loader=None, loader_opts=None):
         loader_f = (lambda *args: _loader(*args, **loader_opts)) if loader_opts else _loader
 
         logging.info("Reading Nodes (METype) info from '%s'", conf.CellLibraryFile)
-        if load_balancer and load_balancer.population != self._target_spec.population:
+        if load_balancer and hasattr(load_balancer, 'population') and load_balancer.population != self._target_spec.population:
```
matz-e, line 204:
```suggestion
        if load_balancer and getattr(load_balancer, 'population', None) != self._target_spec.population:
```

tests/integration-e2e/test_dry_run_worflow.py
```diff
@@ -46,3 +46,31 @@ def test_dry_run_workflow(USECASE3):
...
+    Test that the memory load balance works
+    """
+
+    from neurodamus import Neurodamus
+    nd = Neurodamus(
+        str(USECASE3 / "simulation_sonata.json"),
+        lb_mode="Memory",
+    )
+
+    nd.run()
```
ferdonline, line 76: Maybe we could add some asserts to ensure that the gids are well distributed? It could be nice to do checks on the spikes.

neurodamus/cell_distributor.py
```diff
@@ -201,11 +201,15 @@ def load_nodes(self, load_balancer=None, *, _loader=None, loader_opts=None):
         loader_f = (lambda *args: _loader(*args, **loader_opts)) if loader_opts else _loader
 
         logging.info("Reading Nodes (METype) info from '%s'", conf.CellLibraryFile)
-        if load_balancer and load_balancer.population != self._target_spec.population:
+        if load_balancer and \
```
ferdonline, line 204: For improving the code now I'd create a `MemoryLoadBalance` class that offers a similar API to the old load balance, namely have the `.population` property and related methods, like `get_local_gids(rank)` `valid_load_distribution()`.

ferdonline (commented):
I left some nictpicks for code hygiene, but looking alright!

st4rl3ss:
After testing the current workflow on a variety of circuits it became clear that the current workflow would be pretty clunky for the end user:

1. run a normal dry run to get an estimate for the number of nodes to use
2. multiply that value by the number of physical ranks available on those nodes and re-run a dry run with the `--num-target-ranks` CLI options set to that number in order to generate the allocation file
3. finally run the simulation with `--lb-mode=Memory`

Therefore we paused the approval and review of the MR in order to address this issue, possibly by merging the first two steps into a singular one in order to improve the user experience with the tool.

ferdonline (approved):

PR #131, opened 2024-02-22
[BBPBGLIB-1134] Move ShowProgress.hoc to python

jorblancoa:
## Context
Addresses BBPBGLIB-1134.
Neuron simulation progress hoc file was moved to python.

## Review


neurodamus/core/_utils.py
```diff
@@ -67,8 +68,39 @@ def rank0_wrapper(*args, **kw):
...
+            sec_remain = (self.last_time_check - self.sim_start) * (sim_tstop / sim_t - 1)
+            hours = sec_remain / 3600
+            minutes = (sec_remain % 3600) / 60
+            seconds = (sec_remain % 3600) % 60
+            print(f"\r[t={sim_t:5.2f}] Completed {sim_t*100/sim_tstop:2.0f}% "
+                  f"ETA: {hours:02.0f}:{minutes:02.0f}:{seconds:02.0f}  ", end='', flush=True)
+        Nd.cvode.event(sim_t + 1, self.update_progress)
+
+
+def return_neuron_timings_and_progress(f):
```
matz-e, line 102: I wouldn't rename this - you don't return the progress?

Progress could be a separate decorator, too.

neurodamus/core/_utils.py
```diff
@@ -67,8 +68,47 @@ def rank0_wrapper(*args, **kw):
...
+        Callback function that refreshes the progress value (if enough time has elapsed) and then
+        inserts the next call into the event queue.
+        """
+        current_time = time.time()
+        sim_t = Nd.t
+        sim_tstop = Nd.tstop
+        if (current_time - self.last_time_check > 0.75) and (sim_t > 0):
+            self.last_time_check = current_time
+            sec_remain = (self.last_time_check - self.sim_start) * (sim_tstop / sim_t - 1)
+            hours = sec_remain / 3600
```
mgeplf, line 102: perhaps `timedelta` could be used:
```
In [1]: from datetime import timedelta
In [2]: timedelta(seconds=175823)
Out[2]: datetime.timedelta(days=2, seconds=3023)
In [3]: a=timedelta(seconds=175823)
In [4]: f'{a}'
Out[4]: '2 days, 0:50:23'
```

neurodamus/core/_utils.py
```diff
@@ -67,8 +69,44 @@ def rank0_wrapper(*args, **kw):
         return rank0_wrapper
 
 
+class SimulationProgress:
```
mgeplf, line 72: Does this need to be a decorator? Wouldn't a plain class simplify it? It can be more easily tested, can be instantiated in `_run_neuron`, etc?

A bare function could work, too, by passing `last_time_check`, which would simplify it even more, I think.

jorblancoa: initially i had a simple class instantiated in _run_neuron as you mentioned, but i thought this way was more elegant, and the only difference is the __call__

I could leave the plain class if you prefer

neurodamus/core/_utils.py
```diff
@@ -67,8 +69,36 @@ def rank0_wrapper(*args, **kw):
...
+    def __init__(self):
+        """
+        Class which will set up a timer to perioducally check the amount of time lapsed
+        in the simulation compared to the final tstop value. This is converted into a percentage
+        of the job complete which is then printed to the console.
+        """
+        self.last_time_check = time.time()
+        self.sim_start = self.last_time_check
```
WeinaJi, line 80: In the hoc version, these two vars are initialized for rank 0. Here we do for all ranks. Are they the same ?

neurodamus/node.py
```diff
@@ -1301,8 +1301,7 @@ def run_all(self):
     # -
     @return_neuron_timings
     def _run_neuron(self):
-        Nd.load_hoc("ShowProgress")  # TODO: Drop this
-        _ = Nd.ShowProgress(Nd.cvode, MPI.rank)
+        SimulationProgress()
```
ferdonline, line 1304: IIRC we had to assign to something (hence `_ = ...`) so that the object is not immediately destroyed

neurodamus/core/_utils.py
```diff
@@ -67,8 +69,36 @@ def rank0_wrapper(*args, **kw):
...
+        """
+        current_time = time.time()
+        sim_t = Nd.t
+        sim_tstop = Nd.tstop
+        if (current_time - self.last_time_check > 0.75) and (sim_t > 0):
+            self.last_time_check = current_time
+            sec_remain = (self.last_time_check - self.sim_start) * (sim_tstop / sim_t - 1)
+            print(f"\r[t={sim_t:5.2f}] Completed {sim_t*100/sim_tstop:2.0f}%"
+                  f" ETA: {timedelta(seconds=int(sec_remain))}  ", end='', flush=True)
+        Nd.cvode.event(sim_t + 1, self.update_progress)
```
ferdonline, line 97: Good to know, I wouldn't expect Neuron to accept a class method.

neurodamus/node.py
jorblancoa: It seems to be working like that, maybe because of neuron, but I could do the _ = ... just to be sure

ferdonline (approved):

PR #134, opened 2024-02-27
    [BBPBGLIB-1137] Account for nonzero total current when using SEClamp; allow nonzero total current when simulating physical electrodes

joseph-tharayil:
## Context
NEURON reports SEClamp currents at different time points than all other currents, thus giving inaccurate results in lfp calculations, in cases where the SEClamp represents synaptic inputs. 

This can be fixed by replacing the SEClamp with a new point process mechanism that is identical, except for producing a NONSPECIFIC_CURRENT instead of an ELECTRODE_CURRENT. 

However, in cases where SEClamp or IClamp are used to represent physical electrodes, rather than synaptic inputs, the currents should not be included in the LFP calcualtion

## Scope
When a current_source or conductance_source is used, we check if the alternative mod file, which is expected to be named 'MembraneCurrentSource.mod' or 'ConductanceSource.mod' is available. If so, it is used, with nothing else changed, unless the key-value pair "represents_physical_electrode" is set to True in the Stimulus Config file. Otherwise the IClamp or SEClamp is used as before. 

## Review


neurodamus/core/stimuli.py
```diff
@@ -434,7 +434,21 @@ class _DynamicClamp:
...
+            # Checks if new conductanceSource mechanism is available
+            mt = Neuron.h.MechanismType(1)
+            mname = Neuron.h.ref('')
+            mList = []
+            for i in range(mt.count()):
+                mt.select(i)
+                mt.selected(mname)
+                mList.append(mname[0])
+
+            if 'conductanceSource' in mList:
```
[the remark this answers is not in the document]
joseph-tharayil: Made requested change

neurodamus/stimulus_manager.py
```diff
@@ -752,8 +752,22 @@ def __init__(self, target, stim_info: dict, cell_manager):
...
+                mt = Nd.h.MechanismType(1)
+                mname = Nd.h.ref('')
+                mList = []
+                for i in range(mt.count()):
+                    mt.select(i)
+                    mt.selected(mname)
+                    mList.append(mname[0])
+
+                # If conductanceSource not available, insert standard SEClamp
+                if 'conductanceSource' in mList:
```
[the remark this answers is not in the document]
joseph-tharayil: Fixed

WeinaJi (dismissed):

neurodamus/stimulus_manager.py
```diff
@@ -733,8 +733,13 @@ def __init__(self, target, stim_info: dict, cell_manager):
...
 
-                # create single electrode voltage clamp at location
-                seclamp = Nd.h.SEClamp(tpoint_list.x[sec_id], sec=sc.sec)
+                # If conductanceSource not available, insert standard SEClamp
+                if hasattr(Nd.h, "conductanceSource"):
+                    seclamp = Nd.h.conductanceSource(tpoint_list.x[sec_id], sec=sc.sec)
+                else:
+                    # create single electrode voltage clamp at location
+                    seclamp = Nd.h.SEClamp(tpoint_list.x[sec_id], sec=sc.sec)
+
```
WeinaJi, line 742: The question raised by @joseph-tharayil in https://github.com/BlueBrain/neurodamus-models/pull/6#issuecomment-1994629078 . 

>  a user might want to simulate a "real" SEClamp (i.e., use SEClamp to simulate a physical electrode rather than to account or missing synapses) or even to use it in both ways in the same simulation. In that case the new mod file should not be used. Is there a way to allow the user to specify whether or not they want it?

I am wondering, shall we add a new key in SEClamp to switch between h.conductanceSource and h.SEClamp ? This key will be added in the config file and the SONATA spec for [seclamp](https://sonata-extension.readthedocs.io/en/latest/sonata_simulation.html#seclamp-voltage-clamp)

joseph-tharayil: That seems like a good idea, but having taken a stab at it I think implementing it is a bit beyond my understanding of how neurodamus processes simulation configs. Would someone have the bandwidth this week to help out with that?

WeinaJi: Hi @joseph-tharayil , sorry for my late answer. I can take care of adding the new key in SONATA config and in neurodamus if needed. 
But before that, I am wondering how different are the simulation results between the current `SEClamp` and the new `conductanceSource`. Which mechanim will be the default one in the future for the "seclamp" stimulus?

Invite @seirios and @andrisecker here for the discussion. Thanks!

joseph-tharayil: At least as far as I tested it, the results are identical. The only difference is how the currents are reported. I think the new one should be the default, since my intuition is that users will be more likely to use SEClamp the way we do (to model a missing input) than to model a physical electrode (in which case IClamp would be more appropriate)

joseph-tharayil: Actually, it's now occurring to me that we will run into the same problems with IClamp. Right now, it is included in the LFP calculation, but that should not be the case if it is modeling a physical electrode. Maybe the best option is to include a `use_in_lfp` key in the config file for all current/conductance sources. Then, we can use the new mod file in all cases

joseph-tharayil: Just wanted to follow up on this, since we're almost ready to publish the BlueRecording paper: 

One comment that we got from other scientists was that it might be confusing if the `seclamp` stimulus does not use `h.seclamp`. I think it would make sense if the `conductanceSource` was the default for the `shot_noise` and `ornstein_uhlenbeck` stimuli only, with a key in the config file to switch back to `h.seclamp`. 

Similarly, for current sources, I think the default behavior should be for the current to be included in the LFP calculation, with an optional flag to exclude it.

Does this sound reasonable?

WeinaJi: Hi @joseph-tharayil,
If I understand correctly, for the `class ConductanceSource` in `stimli.py`, we will use `h.conductanceSource` mechenism by default, and a flag from the simulation config file to switch to `h.SEClamp`.  
How about the `class CurrerntSource` ? Is there a change on the currect `h.IClamp` ?

joseph-tharayil: @WeinaJi yes, that's correct regarding `class ConductanceSource`. For `class CurrentSource`, we don't need to make any change to `h.IClamp`. Currently, the `IClamp` current is added to `i_membrane` when the LFP is calculated. This should remain the default behavior, but there should be a flag in the stimulation config to prevent it from being added in the LFP calculation. @jorblancoa would that be something that is easy to implement?

WeinaJi: Hi @joseph-tharayil , after talking to @jorblancoa, I think we need 2 new parameters in the simulation config. 
1. For conductance stimulus type, in the tables of [shot_noise, absolute_shot_noise and relative_shot_noise](https://github.com/BlueBrain/sonata-extension/blob/7208597a9b9eb4423feabbc47c355f85770af64f/source/sonata_simulation.rst?plain=1#L346) and [ornstein_uhlenbeck and relative_ornstein_uhlenbeck ](https://github.com/BlueBrain/sonata-extension/blob/7208597a9b9eb4423feabbc47c355f85770af64f/source/sonata_simulation.rst?plain=1#L373), we need a new string parameter which is either `ConductanceSource` or `SEClamp`. I am not sure how to name it, `conductance_type`? Can you think of a better name?
2. As for the `IClamp` current from LFP, we need a new boolean parameter in the [report table](https://github.com/BlueBrain/sonata-extension/blob/7208597a9b9eb4423feabbc47c355f85770af64f/source/sonata_simulation.rst?plain=1#L402) to exclude `IClamp` from the `lfp` report.
 
Because we are not sure about the names of those two new parameters in the SPEC, is it possible that you create a PR in https://github.com/BlueBrain/sonata-extension to add them?  Then we will take care of implementing it in the simulation toolchain.

joseph-tharayil:
One suggestion we recently got was to add an explicit test that, for a small network, the behavior is identical using the old and the new mod files. I'm not really sure how to implement this. Does anyone have any suggestions?

jorblancoa:
> One suggestion we recently got was to add an explicit test that, for a small network, the behavior is identical using the old and the new mod files. I'm not really sure how to implement this. Does anyone have any suggestions?

Maybe once the new parameter is added a scientific test could be added to run twice with 'represents_physical_electrode' = true and false in order to check the report (and see that the first needs i_membrane+IClamp and the second is the same result only with i_membrane?)

joseph-tharayil:
> > One suggestion we recently got was to add an explicit test that, for a small network, the behavior is identical using the old and the new mod files. I'm not really sure how to implement this. Does anyone have any suggestions?
> 
> Maybe once the new parameter is deployed in libsonata, a scientific test could be added to run a small simulation twice with 'represents_physical_electrode' = true and false in order to check the report (and see that the first with i_membrane+IClamp and the second with only i_membrane gives the same results)

I've created the tests, and they run successfully with `pytest`, but when htey are run through the CI pipeline they cause a segfault. Any idea why this is?

docs/online-lfp.rst
```diff
@@ -130,6 +130,8 @@ Subsequently, an ERROR will be encountered when instantiating the LFP report:
 
 To ensure accurate and valid LFP reports, make sure that the electrodes file corresponds to the circuit being used in your simulation.
 
+- **Stimulus Electrode Compatibility**: A common use case is that current will be injected into a population to account for synaptic inputs from neural populations that are not modeled. In this case, it is neccessary that total current over the neuron sums to zero in order to produce valid extracellular recording results. For IClamp electrodes, this is always the case. However, the Neuron SEClamp class does not fulfill this criterion due to numerical issues. We have created a new point process, `new_conductance_source`, available in `neurodamus-neocortex`, which does fulfill the criterion. Therefore, If an SEClamp source is present in the simulation config file, and `new_conductance_source` is compiled, this will be used instead of the SEClamp mechanism. If `new_conductance_source` is not available, SEClamp will be used, and extracellular recording results should not be trusted.
```
WeinaJi, line 133: Is there update needed for this documentation?
 e.g. it is `ConductanceSource` not `new_conductance_source`. 
 And shall we also mention`MembraneCurrentSource` to replace `IClamp`?

joseph-tharayil: I agree; I've made the changes

WeinaJi (approved):

docs/online-lfp.rst
WeinaJi: Thanks! LGTM!

PR #136, opened 2024-02-29
[BBPBGLIB-1122] Removing legacy RNGs, leaving only Random123

atemerev:
Removing legacy RNGs and the ability to configure RNGs. Random123 is chosen everywhere as the only option.

## Context

RNG types differing from Random123 are no longer used and removed from everywhere (including configuration).

## Scope

RNGSettings.hoc (removed RNGmode and all related constants), configuration.py, random.py, and corresponding usages.

## Testing

Write a test that raises exception in attempt to configure RNG mode.

## Review

### DIFF
--- a/neurodamus/core/random.py
+++ b/neurodamus/core/random.py
@@ -11,7 +11,7 @@ from . import Neuron
 
 class RNG(object):
     def __new__(cls, **kw):
-        """Creates a default RNG, currently based on ACG"""
+        """Creates a default RNG (Random123 with ids of 0,1,2)"""
         seed = kw.get("seed", 0)
         return Random123(0, 1, 2, seed)
 
--- a/neurodamus/core/random.py
+++ b/neurodamus/core/random.py
@@ -10,9 +10,10 @@ from . import Neuron
 
 
 class RNG(object):
-    def __new__(cls):
-        """Creates a default RNG (Random123)"""
-        return Random123(0, 1, 2, 0)
+    def __new__(cls, **kw):
+        """Creates a default RNG, currently based on ACG"""
+        seed = kw.get("seed", 0)
+        return Random123(0, 1, 2, seed)
 
     @classmethod
     def create(cls, ids, seed=None):
--- a/neurodamus/core/random.py
+++ b/neurodamus/core/random.py
@@ -10,10 +10,9 @@ from . import Neuron
 
 
 class RNG(object):
-    def __new__(cls, *ids, **kw):
-        """Creates a default RNG (Random123) """
-        seed = kw.get("seed")
-        return Random123(ids[0], ids[1], ids[2], seed)
+    def __new__(cls):
+        """Creates a default RNG (Random123)"""
+        return Random123(0, 1, 2, 0)
 
     @classmethod
     def create(cls, ids, seed=None):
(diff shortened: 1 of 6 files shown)


neurodamus/data/hoc/RNGSettings.hoc
WeinaJi, line 1: Can we replace the hoc template with the python class`RNGConfig` in `configuration.py`?

WeinaJi: I see this hoc template is called by our synapse helper functions, such as https://github.com/BlueBrain/neurodamus-models/blob/main/common/hoc/AMPANMDAHelper.hoc . It is pity that the call is just to get a `SynapseSeed` which is set globally. We could pass the seed to the helper init function instead. Will leave the clean up later on.

neurodamus/core/random.py
```diff
@@ -3,57 +3,36 @@
...
 #       Neuron objects are returned (using __new__)
 
 
 class RNG(object):
     def __new__(cls, *ids, **kw):
-        """Creates a default RNG, currently based on ACG"""
+        """Creates a default RNG (Random123) """
         seed = kw.get("seed")
-        return Neuron.h.Random() if seed is None else Neuron.h.Random(seed)
+        return Random123(ids[0], ids[1], ids[2], seed)
```
WeinaJi, line 16: A nitpick, shall we also assert len(ids) ?

atemerev: I think I had it the other way around, RNG() should create some default instance of a random generator (I am not sure this is a good idea, but it is used throughout the code, e.g. in `stimuli.py`). We can do it with seed=0 (deterministic) or seed=current time (stochastic), I think deterministic is OK for now.

neurodamus/core/random.py
```diff
@@ -3,57 +3,36 @@
...
 
-# NOTE: These are pure wrappers, in the sense they dont create Python objects. Instead
+# NOTE: These are pure wrappers, in the sense they don't create Python objects. Instead
 #       Neuron objects are returned (using __new__)
 
 
 class RNG(object):
-    def __new__(cls, *ids, **kw):
+    def __new__(cls, **kw):
         """Creates a default RNG, currently based on ACG"""
```
WeinaJi, line 14: Just discussed with @alkino  and @pramodk , it looks good that we instantiate Random123 object with a (0,1,2, seed), the first 3 seeds will be changed later on during synapse creation by gid, syn_id, etc.

Another nitpick, could you please change the docstring to replace ACG with Random123 ?

neurodamus/data/hoc/RNGSettings.hoc
```diff
@@ -17,15 +17,13 @@ globalSeed = 0
...
 public init, interpret, getRNGMode, getIonChannelSeed, getSynapseSeed, getStimulusSeed, getMinisSeed, getGlobalSeed
-public COMPATIBILITY, RANDOM123, UPMCELLRAN4
+public RANDOM123
 
 external rngMode, ionchannelSeed, synapseSeed, stimulusSeed, minisSeed, globalSeed, terminate
 
 proc init() {
     // consts for random number handling
-    COMPATIBILITY = 0
     RANDOM123 = 1
```
WeinaJi, line 26: Can we also remove `rngMode` and `getRNGMode` ?  I see that you have removed them in`connection.py` and `stimulus_manager.py`.

atemerev: OK, apparently, their usages were removed in neurodamus-models, so I can do that.

WeinaJi (approved):

PR #144, opened 2024-03-19
Clean up wheel contents.

### LINKED ISSUE #143: Wheel pollutes Python's site packages directory
With `pip install neurodamus`, I would expect to find all relevant files in `.../site-packages/neurodamus`. Yet `pip list` tells me that I also installed

* `_benchmarks`
* `ci`
* `core/{hoc,mod}`
* `docker`
* `examples`

All these should not really be present. For the HOC, MOD files, maybe we should move them into `neurodamus/data/{hoc,mod}`?

matz-e:
Fixes #143. Replaces `setuptools`, which seems very happy to just add
everything in the current directory into the source distribution with
`hatch`, which is much, much, much more reasonable. Also:

* Delete now obsolete `MANIFEST.in`
* Delete outdated, non-functional `_debug.py`
* Move `init.py`, `core/hoc`, `core/mod`  into `neurodamus/data` to properly compartmentalize it


WeinaJi:
Thanks for the improvements. I've learned new things about `hatching`.
Just a question: I ran `hatch version` on a local checkout of  tag `3.1.1` and the version returned is `3.1.2.dev0` instead of `3.1.1` . Will we get the same kind of version on the wheel generated at new tags?

matz-e:
> Thanks for the improvements. I've learned new things about `hatching`. Just a question: I ran `hatch version` on a local checkout of tag `3.1.1` and the version returned is `3.1.2.dev0` instead of `3.1.1` . Will we get the same kind of version on the wheel generated at new tags?

Hm, it should. If I checkout this PR, run `git tag 123` and then `hatchling version`, I get the expected output of `123`. Do you have steps I can reproduce?

WeinaJi:
> > Thanks for the improvements. I've learned new things about `hatching`. Just a question: I ran `hatch version` on a local checkout of tag `3.1.1` and the version returned is `3.1.2.dev0` instead of `3.1.1` . Will we get the same kind of version on the wheel generated at new tags?
> 
> Hm, it should. If I checkout this PR, run `git tag 123` and then `hatchling version`, I get the expected output of `123`. Do you have steps I can reproduce?

Ah yes, you are right. In fact, I made a change in my local check out of the tag, that's why I got dev0. The versioning is ok.

WeinaJi (approved):

WeinaJi:
I also tag @ferdonline @jorblancoa @atemerev here, to inform them of the new location of the `hoc`/`mod` folders  and `init.py` file after this PR is merged.

PR #145, opened 2024-03-20
[BBPBGLIB-1057] Replace Report.hoc by report.py

jorblancoa:
## Context
Create a python report.py and replace the old Report.hoc

## Review


neurodamus/node.py
```diff
@@ -854,7 +854,8 @@ def enable_reports(self):
             if SimConfig.restore_coreneuron:
                 continue  # we dont even need to initialize reports
 
-            report = Report(*rep_params)
+            has_gids = self._circuits.global_manager.get_final_gids().size > 0
```
matz-e, line 857: We don't have a `len()` implementation for that?

neurodamus/node.py
```diff
@@ -933,7 +937,7 @@ def _report_build_params(self, rep_name, rep_conf, target, pop_offsets_alias_pop
             end_time,
             SimConfig.output_root,
             None,
-            Nd.String(rep_conf["Scaling"]) if "Scaling" in rep_conf else None,
+            rep_conf["Scaling"] if "Scaling" in rep_conf else None,
```
matz-e, line 940:
```suggestion
            rep_conf.get("Scaling"),
```
Assuming this is just a `dict`?

neurodamus/node.py
```diff
@@ -853,10 +854,13 @@ def enable_reports(self):
...
 
-            report = Nd.Report(*rep_params)
+            has_gids = self._circuits.global_manager.get_final_gids().size > 0
+            report = Report(*rep_params) if has_gids else None
 
             if not SimConfig.use_coreneuron or rep_params.rep_type == "Synapse":
-                self._report_setup(report, rep_conf, target, rep_params.rep_type)
+                if not self._report_setup(report, rep_conf, target, rep_params.rep_type):
+                    n_errors += 1
+                    continue
```
matz-e, line 863: Instead of having `_report_setup` return a `bool` and catch any exception, you could do the exception catching here. The one early `return False` could also `raise`.

Would sort of force exception/error treatment in the caller (here), which seems a bit more appropriate, and it makes it harder to forget checking the error return value.

neurodamus/node.py
```diff
@@ -933,7 +937,7 @@ def _report_build_params(self, rep_name, rep_conf, target, pop_offsets_alias_pop
             end_time,
             SimConfig.output_root,
```
matz-e, line 938: This could also contain `SimConfig.use_coreneuron` (see comment below)

neurodamus/report.py
```diff
@@ -0,0 +1,259 @@
...
+        if scaling_option is None or scaling_option == 'Area':
+            return 1  # SCALING_AREA
+        elif scaling_option == 'None':
+            return 0  # SCALING_NONE
+        else:
+            return 2  # SCALING_ELECTRODE
+
+    def add_compartment_report(self, cell_obj, point, vgid, use_coreneuron=False,
+                                pop_name="default", pop_offset=0):
+        if use_coreneuron:
```
matz-e, line 68: Can this be an object setting, rather than being queried / passed through for every cell?

matz-e (commented):
The less HOC the better :partying_face:

neurodamus/node.py
jorblancoa: the idea was to show the errors in all reports before raising the exception, that way users can fix all problems and not going one by one...

matz-e: I meant something along
```
try:
    self._report_setup(...)
except:
    n_errors += 1
```
to avoid anyone not ever checking the return value of `_report_setup`

neurodamus/target_manager.py
```diff
@@ -622,3 +622,43 @@ def __init__(self, cell):
                 # Store a SectionRef to the section at the index specified by v_value
                 self.isec2sec[int(v_value)] = Nd.SectionRef(sec=sec)
             index += 1
+
+
+class TPointList:
```
matz-e, line 627: I hope that at some point (soon?) we could replace this with a more Pythonic construct.

matz-e (approved):

ferdonline:
Great change! My only slight concern is regarding performance since reports can demanding as they go through all sections and hoc can use optimized 'for' loops, like `forsec`. I know it's tedious, but could you have maybe a simple time measurement, showing that we don't have a regression in terms of report setting-up time?

jorblancoa:
> Great change! My only slight concern is regarding performance since reports can demanding as they go through all sections and hoc can use optimized 'for' loops, like `forsec`. I know it's tedious, but could you have maybe a simple time measurement, showing that we don't have a regression in terms of report setting-up time?

Actually we have the times of the report setup time. It is true that is slower, but do we care that much? it is usually insignificant compared to the run time.

I checked the blueconfigs sonataconf-quick-v5-plasticity that has 5 reports, and run a simulation for 10ms and changed the target to Mosaic (~220k cells and 110 gids per rank), so this is almost worst case scenario (short simulation with a lot of reports and big number of gids per rank).

The hoc version:
```
[INFO] +================================================ TIMEIT STATS =================================================+
[INFO] |                       Event Label                        | Avg.Time | Min.Time | Max.Time | Hits R0 / Total   |
[INFO] +---------------------------------------------------------------------------------------------------------------+
[INFO] | Compute LB                                               |     0.02 |     0.01 |     0.03 |       1 / 2.00K   |
[INFO] | Cell creation                                            |    44.23 |    44.23 |    44.23 |       1 / 2.00K   |
[INFO] | Synapse creation                                         |    43.57 |    43.57 |    43.57 |       1 / 2.00K   |
[INFO] | Replay init                                              |     0.05 |     0.03 |     0.06 |       3 / 6.00K   |
[INFO] | Replay inject                                            |     0.16 |     0.14 |     0.17 |       1 / 2.00K   |
[INFO] | Enable Stimulus                                          |     1.16 |     1.12 |     1.68 |       1 / 2.00K   |
[INFO] | Enable Modifications                                     |     0.00 |     0.00 |     0.00 |       1 / 2.00K   |
[INFO] | Enable Reports                                           |    18.81 |    18.59 |    19.03 |       1 / 2.00K   |
[INFO] | Model Finalized                                          |    14.30 |    14.30 |    14.30 |       1 / 2.00K   |
[INFO] |  ╚ stdinit                                               |    14.25 |    14.25 |    14.25 |       1 / 2.00K   |
[INFO] | finished Run                                             |   148.49 |   148.49 |   148.49 |       1 / 2.00K   |
[INFO] |  ╚ psolve                                                |   146.57 |   146.57 |   146.60 |       1 / 2.00K   |
[INFO] +---------------------------------------------------------------------------------------------------------------+
```

And the python version:
```
[INFO] +================================================ TIMEIT STATS =================================================+
[INFO] |                       Event Label                        | Avg.Time | Min.Time | Max.Time | Hits R0 / Total   |
[INFO] +---------------------------------------------------------------------------------------------------------------+
[INFO] | Compute LB                                               |     0.02 |     0.01 |     0.03 |       1 / 2.00K   |
[INFO] | Cell creation                                            |    44.52 |    44.52 |    44.52 |       1 / 2.00K   |
[INFO] | Synapse creation                                         |    40.99 |    40.99 |    40.99 |       1 / 2.00K   |
[INFO] | Replay init                                              |     0.04 |     0.03 |     0.05 |       3 / 6.00K   |
[INFO] | Replay inject                                            |     0.16 |     0.15 |     0.18 |       1 / 2.00K   |
[INFO] | Enable Stimulus                                          |     1.16 |     1.12 |     1.58 |       1 / 2.00K   |
[INFO] | Enable Modifications                                     |     0.00 |     0.00 |     0.00 |       1 / 2.00K   |
[INFO] | Enable Reports                                           |    31.88 |    31.64 |    32.09 |       1 / 2.00K   |
[INFO] | Model Finalized                                          |    13.75 |    13.75 |    13.75 |       1 / 2.00K   |
[INFO] |  ╚ stdinit                                               |    13.71 |    13.71 |    13.71 |       1 / 2.00K   |
[INFO] | finished Run                                             |   144.41 |   144.41 |   144.41 |       1 / 2.00K   |
[INFO] |  ╚ psolve                                                |   143.83 |   143.83 |   143.85 |       1 / 2.00K   |
[INFO] +---------------------------------------------------------------------------------------------------------------+
```


What do you think?

ferdonline:
Thanks for the measurements. I think that's acceptable given the benefit. 👍

WeinaJi (approved):

WeinaJi (dismissed):

WeinaJi (approved):

PR #146, opened 2024-03-21
[BBPBGLIB-1139] config / model error logging fix, stage 2

atemerev:
## Context

On simulation launch in commands.py, exceptions were handled in the same way for configuration / model loading errors, and simulation errors. This required quirky synchronization to make sure that exceptions were logged only at a single MPI node, otherwise they flood the output.

The idea of this PR is to separate exception handling for configuration parsing / model loading (these errors are the same for all nodes, and supposed to be logged only at the master node), and simulation errors (can happen only at some runs, and can be different everywhere, and perhaps need to be logged at all nodes).

## Scope

Separate exception handling at model loading stage and simulation run stage. Call _mpi_abort only in the latter case. In the former case, log errors only on the node with the MPI rank 0.

## Testing

Again, I don't think it is feasible to write a unit test for this.

## Review
* [X] Updated Readme, in-code, developer documentation

### DIFF
--- a/neurodamus/commands.py
+++ b/neurodamus/commands.py
@@ -62,6 +62,7 @@ def neurodamus(args=None):
     options = docopt_sanitize(docopt(neurodamus.__doc__, args))
     config_file = options.pop("ConfigFile")
     log_level = _pop_log_level(options)
+    ALL_RANKS_SYNC_WINDOW = 1
 
     if not os.path.isfile(config_file):
         logging.error("Config file not found: %s", config_file)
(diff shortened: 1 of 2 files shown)


WeinaJi:
Hi @atemerev , there is the exception type `ConfigurationError` which should be raised for errors during reading config files. This exception should be raised by all ranks, and to be caught properly. The errors during modelling, such as read and creation of cells and synapses are more complex. They may happen in some of the ranks but not all. An example what I can think of is loading the emodel hoc template in [Cell_V6._instantiate_cell](https://github.com/BlueBrain/neurodamus/blob/20be81f81853a7db6f10fa4754dca59fa16412c5/neurodamus/metype.py#L164) where the `EModel` files are from scientists and some of them may contain errors. As difference cells require different EModel templates, the `EModel` files load in each rank are not the same meaning that we may have errors in some of the ranks. In this case, we may not be able to log it at rank 0.

neurodamus/commands.py
```diff
@@ -74,22 +75,25 @@ def neurodamus(args=None):
...
-        return 1  # no need for _mpi_abort, error is being handled by all ranks
-    except Exception:
-        show_exception_abort("Unhandled Exception. Terminating...", sys.exc_info())
+
+    try:
+        nrdm.run()
+    except Exception as e:
+        # at this stage, this is an error in the simulation itself, can happen in individual nodes
+        # it is OK to log it everywhere it happens
+        logging.critical(f"Unhandled exception. Terminating: {str(e)}", sys.exc_info())
```
matz-e, line 92: Can we use `logging.exception`, would that make sense?

WeinaJi:
After the final decision addressed on BBPBGLIB-1139, we can close this PR.

PR #148, opened 2024-04-04
[BBPBGLIB-1145] Write gids without offset in WholeCell LB complexity file

jorblancoa:
## Context
Addresses issue highlighted in BBPBGLIB-1145, where the offsets were being calculated before the load balancing, causing issue sometimes when having virtual populations alphabetically ordered before the first real population.

## Review


jorblancoa:
I tested locally the SSCx simulation that had the issue and also launched the long tests, to be sure we don't have the problem found in the past:
https://bbpgitlab.epfl.ch/hpc/sim/blueconfigs/-/jobs/1229037

jorblancoa:
Long tests are passing also!
https://bbpgitlab.epfl.ch/hpc/sim/blueconfigs/-/pipelines/203645

WeinaJi (dismissed):
Looks good ! Thank you!

Do we need a unit test for this ? I let you decide.

tests/integration-e2e/test_loadbalance.py
```diff
@@ -140,18 +140,81 @@ def test_multisplit(target_manager, circuit_conf, capsys):
...
+    nd = Node(tmp_file.name, {"lb_mode": "WholeCell"})
     nd.load_targets()
+    SimConfig.check_connections_configure(nd._target_manager)
     lb = nd.compute_load_balance()
     nd.create_cells(lb)
 
+    # Check the complexity file
+    base_dir = "sim_conf"
+    pattern = "_loadbal_*.*"  # Matches any hash and population
+    cx_pattern = "cx_*#.dat"  # Matches any cx file with the pattern
```
WeinaJi, line 210: Can we hard code the population name in the pattern? I assume the cx files are produced for `NodeA`?

tests/integration-e2e/test_loadbalance.py
```diff
@@ -140,18 +140,81 @@ def test_multisplit(target_manager, circuit_conf, capsys):
...
+    # Check the complexity file
+    base_dir = "sim_conf"
+    pattern = "_loadbal_*.NodeA"  # Matches any hash and population
+    cx_pattern = "cx_NodeA*#.dat"  # Matches any cx file with the pattern
+    assert Path(base_dir).is_dir(), "Directory 'sim_conf' not found."
+    cx_file = _read_complexity_file(base_dir, pattern, cx_pattern)
+    lines = cx_file.splitlines()
+    assert int(lines[1]) == 3, "Number of gids different than 3."
+    # Gid should be without offset (2 instead of 1002)
+    assert int(lines[3].split()[0]) == 2, "gid 2 not found."
```
ferdonline, line 216: Nice test

neurodamus/cell_distributor.py
```diff
@@ -502,7 +506,7 @@ def getSpGid(self, gid):
         """
         manager = self._find_manager(gid)
         if manager._binfo:
-            return manager._binfo.thishost_gid(gid)
+            return manager._binfo.thishost_gid(gid - manager.local_nodes.offset)
```
ferdonline, line 509: I wonder why we don't do `+ gid_offset` like 10 lines above. I don't racall if this should do the same, just noticing it seems inconsistent and if you know the reason for the difference maybe let's write a comment

jorblancoa: You are right, I probably forgot because I think this function is not used...

WeinaJi (approved):

PR #151, opened 2024-04-09
[BBPGLIB-1148] Add support for gap junctions in dry run

st4rl3ss:
## Context
This MR solves https://bbpteam.epfl.ch/project/issues/browse/BBPBGLIB-1148

After investigation it seems that the `sonataconf-thalamus` circuit from `blueconfig` is including also gap junctions, that's why some of the connections were missing the `syn_type_id`.
Now the dry run can correctly treat gap junctions, count them and add to the synapse memory total. All the other types are also counted for future use.

## Scope
Made the `get_counts` function work also when no `group_by` is specified, in which case it will return an int.
Added more branch and checks to take into account non-chemical connections.

## Testing
No further testing needed.

## Review

### DIFF
--- a/neurodamus/connection_manager.py
+++ b/neurodamus/connection_manager.py
@@ -9,6 +9,7 @@ from collections import defaultdict, Counter
 from itertools import chain
 from os import path as ospath
 from typing import List, Optional
+from libsonata._libsonata import SonataError
 
 from .core import NeurodamusCore as Nd
 from .core import ProgressBarRank0 as ProgressBar, MPI
@@ -775,7 +776,13 @@ class ConnectionManagerBase(object):
                 sample_len = len(sample)
                 if not sample_len:
                     continue
-                sample_counts = self._synapse_reader.get_counts(sample, group_by="syn_type_id")
+                try:
+                    sample_counts = self._synapse_reader.get_counts(sample, group_by="syn_type_id")
+                except SonataError as e:
+                    logging.warning("Error while getting synapse counts: %s", e)
+                    logging.warning("You might be using a non-compliant version of the edge file.")
+                    continue
+
                 logging.debug("Gids: %s... Types: %s", sample[:10], sample_counts)
                 logging.debug("Average syn/cell: %.2f", sum(sample_counts.values()) / sample_len)
                 sampled_gids_count += sample_len
--- a/neurodamus/connection_manager.py
+++ b/neurodamus/connection_manager.py
@@ -9,7 +9,7 @@ from collections import defaultdict, Counter
 from itertools import chain
 from os import path as ospath
 from typing import List, Optional
-from libsonata._libsonata import SonataError
+from libsonata import SonataError
 
 from .core import NeurodamusCore as Nd
 from .core import ProgressBarRank0 as ProgressBar, MPI
--- a/neurodamus/connection_manager.py
+++ b/neurodamus/connection_manager.py
@@ -777,12 +777,16 @@ class ConnectionManagerBase(object):
                 if not sample_len:
                     continue
                 try:
-                    sample_counts = self._synapse_reader.get_counts(sample, group_by="syn_type_id")
+                    if self.CONNECTIONS_TYPE == ConnectionTypes.Synaptic:
+                        sample_counts = self._synapse_reader.get_counts(
+                            sample, group_by="syn_type_id")
+                    else:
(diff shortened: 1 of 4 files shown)


neurodamus/connection_manager.py
```diff
@@ -775,7 +776,13 @@ def _get_conn_stats(self, dst_target):
                 sample_len = len(sample)
                 if not sample_len:
                     continue
-                sample_counts = self._synapse_reader.get_counts(sample, group_by="syn_type_id")
+                try:
+                    sample_counts = self._synapse_reader.get_counts(sample, group_by="syn_type_id")
```
WeinaJi, line 780: `syn_type_id` is mandatory for "chemical" edges. But it is not required for other type of edges file, e.g. "electrical", the ngv edges. See the [SONATA SPEC](https://sonata-extension.readthedocs.io/en/latest/sonata_tech.html#edge-file). Is it possible to count with another (maybe common) attributes?

neurodamus/connection_manager.py
```diff
@@ -9,6 +9,7 @@
 from itertools import chain
 from os import path as ospath
 from typing import List, Optional
+from libsonata._libsonata import SonataError
```
mgeplf, line 12: importing from private (ie: leading underscore) parts of packages is not generally recommended.

The following should work:
`from libsonata import SonataError`

neurodamus/connection_manager.py
st4rl3ss: Thanks for the comment Weina. 
The field is used for counting but also to discriminate between different types of synapses.
I guess that in the case of electrical synapses there's no need to count different types?

WeinaJi: You are right, for electrical synapses, there is only one type : the gap juction `Nd.Gap` .

st4rl3ss: So we can make the assumption that if there's no `syn_type_id` we should count them as `Nd.Gap`?

WeinaJi: Maybe do like this: if calling `_get_conn_stats` from  `GapJunctionManager` class, we count all synapse as `Nd.Gap`. 
In the ngv case, the `astrocyte` edges don't  contain  the  attribute `syn_type_id` either, but there the creation of syn objects is different from the general case.

neurodamus/utils/memory.py
```diff
@@ -309,24 +311,45 @@ def collect_display_syn_counts(self):
...
         log_verbose("+{:=^68}+".format(" Synapse Count "))
         log_verbose("| {:^40s} | {:^10s} | {:^10s} |".format("Synapse Type", "Family", "Count"))
         log_verbose("+{:-^68}+".format(""))
-        for syn_type, count in sorted(master_counter.items()):
+
+        # Some synapse types are numeric, others are strings, so we need to handle both
+        numeric_items = [(syn_type, count)
+                         for syn_type, count in master_counter.items()
+                         if isinstance(syn_type, (int, np.integer))]
+        for syn_type, count in sorted(numeric_items):
```
WeinaJi, line 323: Could you remind me why we sort "sorted(numeric_items)"?

st4rl3ss: It's just for visualization, so when we print the table containing all the different types of synapses and their respective memory footprint they are ordered by their `syn_type_id`.

WeinaJi: If it is tolerant to print without sorting, I think we can fill the `inh/exc/gap/other_count` using one `for-loop`.  What is the usual size of `master_counter` ?

st4rl3ss: It really depends on the circuit but normally we can expect in the most diverse circuits something in the order of ~40-50 different `syn_type_ids` i.e. `master_counter` keys to order. So the sorting is no big deal normally but if you think it's useless to order the synapses for visualization I can remove the sorting and make just one for loop, also removing the need for numeric vs string synapse types.

WeinaJi: OK, 40-50 is not a big size for loop, sort and then loop. We can stay like this ATM I think. 

BTW, I just pushed a commit to complete the `CONNECTION_TYPE` for NeuroModulationManager for the dry-run counting.

jorblancoa (approved):
LGTM!

WeinaJi (approved):

PR #155, opened 2024-04-18
Allow API configuration to preserve the axons during setup

ferdonline:
## Context
With BlueConfig one could ask a circuit to preserve the axons in full detail via a config entry `DetailedAxons`.

However, that options was never made globally available via a flag, which enables it also for Sonata simulations.

This PR covers that.

## Scope
 - Introduce the new Neurodamus top-level option `keep_axon`
 - Propagate to all Detailed neuron circuits which don't set the option explicitly.

## Testing
in the testing environment under `test/ngv_wip` with the patched `emodels_atp_1p4` one must see

**Script:**
```python
nd = neurodamus.Neurodamus(
    "simulation_config.json",
    keep_axon=True,
    auto_init=False,
    cleanup_atexit=False,
    logging_level=2,
)
nd.init()
nd.run()
```
**Output:**
```
[VERB]  -> Keeping axons ENABLED
...
[INFO] Executing actions after stdinit...
[INFO] Now deleting the axon!
REPLACING AXON!
[INFO] Now deleting the axon!
REPLACING AXON!
```

## Review


cattabiani:
I wrote here my findings:

https://bbpteam.epfl.ch/project/issues/browse/BBPBGLIB-984.

Let me know if you prefer the reports directly here instead

WeinaJi (approved):
LGTM!
Just one question: is there any need to add a CLI option for this parameter?

cattabiani:
not from me

jorblancoa (approved):

WeinaJi (approved):

PR #165, opened 2024-04-30
[BBPBGLIB-1158] Add empirical mechanism to estimate simulation memory usage

st4rl3ss:
## Context
Add an estimate for the simulation memory usage. The estimate is simply calculated considering 2.5 times the total amount of memory estimated for cell + synapses. 
This factor is based on tests that we performed on a variety of circuit that seem to suggest that the ratio between the instantiation and the simulation tops at 2.5 times.

For more information see: https://bbpteam.epfl.ch/project/issues/browse/BBPBGLIB-1158

## Scope
Just a simple add to already existing functions.

## Testing
The modification will be tested during the already implemented integration tests for the dry run workflow.

## Review

### DIFF
--- a/neurodamus/utils/memory.py
+++ b/neurodamus/utils/memory.py
@@ -22,6 +22,11 @@ from ..io.sonata_config import ConnectionTypes
 
 import numpy as np
 
+# The factor to multiply the cell + synapses memory usage by to get the simulation memory estimate.
+# This is an heuristic estimate based on tests on multiple circuits.
+# More info in docs/architecture.rst.
+SIM_ESTIMATE_FACTOR = 2.5
+
 
 def trim_memory():
     """
@@ -256,7 +261,6 @@ class DryRunStats:
         self.metype_counts = Counter()
         self.synapse_counts = Counter()
         self.suggested_nodes = 0
-        self.sim_estimate_factor = 2.5
         _, _, self.base_memory, _ = get_task_level_mem_usage()
 
     @run_only_rank0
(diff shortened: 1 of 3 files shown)


neurodamus/utils/memory.py
```diff
@@ -256,6 +256,7 @@ def __init__(self) -> None:
         self.metype_counts = Counter()
         self.synapse_counts = Counter()
         self.suggested_nodes = 0
+        self.sim_estimate_factor = 2.5
```
matz-e, line 259: This could also be an all caps file wide variable, with a comment on why 2.5 and how it's going to bu used :)

st4rl3ss: Fair enough! Fill fix, thanks!

neurodamus/utils/memory.py
```diff
@@ -361,8 +366,14 @@ def display_total(self):
         logging.info("| {:<40s} | {:12.1f} |".format(f"Overhead (ranks={MPI.size})", full_overhead))
         logging.info("| {:<40s} | {:12.1f} |".format("Cells", self.cell_memory_total))
         logging.info("| {:<40s} | {:12.1f} |".format("Synapses", self.synapse_memory_total))
+        self.simulation_estimate = (self.cell_memory_total
```
ferdonline, line 369: I had understood that 2.5x was the factor among the totals. E.g. if we had estimated using 10 we were seeing 25 being used during simulation. If that's the case 25 should me considered the grand total, not exclusively the simulation, and hence not summed to the base 10.

ferdonline (approved):
I left a comment just to clarify a point that I maybe got wrong. If that's the case please ignore and feel free to merge

neurodamus/utils/memory.py
st4rl3ss: I double checked just to make sure about it and no, the simulation memory usage data I took is separate from the cell + synapse instantiation and therefore should be added to it and does not represent the total.

By re-running the test for this purposes I noticed a significantly smaller memory footprint for the simulation than what we've seen in the past, possibly due to memory optimization of the last months.

I will leave the 2.5x factor for now but we might want to consider to lower it a bit in the future.

PR #168, opened 2024-05-08
Avoid calling cleanup from destructor

ferdonline:
## Context
Neurodamus used to ensure `cleanup` was called by calling it from the destructor of the `Neurodamus` class.

However, depending on the python version, this could lead to crashes as some objects from `Node` eventually were already destroyed.

To workaround this problem, at some point we introduced the `cleanup_atexit` parameter which would skip over this procedure, and notably multiscale_run was using it. However this was not a solution as cleanup was not being called.

## Scope
This PR fixes this long standing issue. Cleanup is now called directly from the top-level `run()` function. It can still be avoided if the user really wants to keep the data, with the parameter `cleanup=False`
 
## Testing
CI is happy.

## Review

### DIFF
--- a/neurodamus/commands.py
+++ b/neurodamus/commands.py
@@ -11,6 +11,8 @@ from docopt import docopt
 from os.path import abspath
 from pathlib import Path
 
+from neurodamus.utils.timeit import TimerManager
+
 from . import Neurodamus
 from .core import MPI, OtherRankError
 from .core.configuration import ConfigurationError, LogLevel, EXCEPTION_NODE_FILENAME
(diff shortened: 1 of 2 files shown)


WeinaJi (approved):
LGTM! Thanks!
@cattabiani Once this change is deployed, in multiscale_run you will have to remove `cleanup_atexit` from calling neurodamus.

WeinaJi:
Shall we merge this PR ?

jorblancoa:
> Shall we merge this PR ?

Looks good to me!

jorblancoa (approved):

ferdonline:
Tests just failed because pytest-forked depends on py.process which went away. https://github.com/pytest-dev/pytest-forked/issues/88

jorblancoa (approved):

PR #171, opened 2024-05-13
[BBPBGLIB-1152] Multi-cycle support for dry run estimate and instantiation

st4rl3ss:
## Context
Addresses https://bbpteam.epfl.ch/project/issues/browse/BBPBGLIB-1152 by saving the weight estimate data per cell to file , distributing cells among cycles of a multi-cycle run when `modelbuilding_steps` is defined and instantiating multi-cycle allocation from file.

The allocation file structures change from a simple dict of a dict where the first key is the population and the second one the rank, to a slightly more complex one where the second key is now a tuple made of assigned rank and cycle respectively.

## Scope
No new function are added for the moment, just a rework and expansion of the present ones.

The MR tries to reuse as much as possible the existing multi-cycle pipeline and void complete reworking. 

## Testing
Added tests for multi-cycle estimate, allocation and instantiation.
Modified existing tests to accommodate new allocation file structure. 

## Review


neurodamus/utils/memory.py
```diff
@@ -513,8 +539,33 @@ def assign_cells_to_rank(rank_allocation, rank_memory, batch, batch_memory):
             all_allocation[pop] = rank_allocation
             all_memory[pop] = rank_memory
 
-        print_allocation_stats(all_allocation, all_memory)
-        export_allocation_stats(all_allocation, self._ALLOCATION_FILENAME)
+        # If cycles is not None and more than 1, distribute cells also over cycles
+        if cycles is not None and cycles > 1:
```
matz-e, line 544: `if cycles:` should be enough in that case, no?

st4rl3ss: Mostly yes but in case someone haphazardly sets `modelbuilding_steps =1` the code executed would be basically redundant distributing cells over 1 cycle. Not a big deal TBH but also not a big deal to require a strict `cycles > 1` also for better clarity of the code.

st4rl3ss: NVM, I've reworked that part to solve other issues I had with the shapes of the resulting dictionaries. Thanks a lot for pointing this out.

docs/examples.rst
```diff
@@ -90,25 +90,60 @@ for SONATA circuits.
...
+cycles. For example:
+
+``neurodamus --configFile=BlueConfig --dry-run --modelbuilding-steps=10``
+
+This will generate the same files as the previous example, just distributed along ranks and cycles.
+These files can then be used by running the simulation with the following flags:
+
+``neurodamus --configFile=BlueConfig --lb-mode=Memory --modelbuilding-steps=10``
+
+This can further improve optimize the memory usage of the simulation and avoid OOM errors.
```
matz-e, line 146: Should we really be using `BlueConfig` still?

neurodamus/utils/memory.py
```diff
@@ -198,25 +198,31 @@ def print_allocation_stats(rank_allocation, rank_memory):
...
     """
-    Export allocation dictionary to serialized pickle file.
+    Export allocation dictionary to a serialized pickle file and metype memory usage to a JSON file.
     """
+    # Export rank_allocation to a compressed pickle file
     compressed_data = gzip.compress(pickle.dumps(rank_allocation))
     with open(filename, 'wb') as f:
         f.write(compressed_data)
 
+    # Export metype_memory_usage to a JSON file
```
matz-e, line 210:
```suggestion
    compressed_data = gzip.compress(pickle.dumps(rank_allocation))
    with open(filename, 'wb') as f:
        f.write(compressed_data)

```
The comments just state what the code tells you without further informational content.

neurodamus/utils/memory.py
```diff
@@ -459,54 +474,62 @@ def get_num_target_ranks(self, num_ranks):
...
-                                and values are the total memory load on each rank.
+            bucket_allocation (dict): A dictionary where keys are tuples (pop, rank_id, cycle_id)
+                                    and values are lists of cell IDs assigned to each bucket.
+            bucket_memory (dict): A dictionary where keys are tuples (pop, rank_id, cycle_id)
+                                and values are the total memory load on each bucket.
+            metype_memory_usage (dict): A dictionary where keys are METype IDs
+                                        and values are the memory load of each METype.
         """
-        logging.info("Distributing cells across %d ranks", num_ranks)
+        if cycles is None or cycles <= 1:
```
matz-e, line 497:
```suggestion
        if not cycles:
```
unless there's a real chance that someone will use negative numbers here.

matz-e (commented):
LGTM. Minor comments.

neurodamus/utils/memory.py
st4rl3ss: Since the same function is exporting two different files in two differents format I thought they might be informative. But the code is pretty much self-explanatory. 
Will remove the comments.

neurodamus/node.py
```diff
@@ -375,14 +375,14 @@ def compute_load_balance(self):
...
-            alloc = import_allocation_stats("allocation.pkl.gz")
+            alloc = import_allocation_stats("allocation.pkl.gz", self._cycle_i)
             for pop, ranks in alloc.items():
                 for rank, gids in ranks.items():
                     logging.debug(f"Population: {pop}, Rank: {rank}, Number of GIDs: {len(gids)}")
             if MPI.rank == 0:
                 unique_ranks = set(rank for pop in alloc.values() for rank in pop.keys())
                 logging.debug("Unique ranks in allocation file: %s", len(unique_ranks))
-                if MPI.size != len(unique_ranks):
+                if MPI.size < len(unique_ranks):
```
ferdonline, line 385: Do we want to proceed knowing the resources will be under-utilized?
Either way, since we store the information about METypes we can redistribute cells very quickly. So are we leaving that for the next iteration and for the moment we simply error out?

neurodamus/utils/memory.py
```diff
@@ -198,25 +198,29 @@ def print_allocation_stats(rank_allocation, rank_memory):
 
 
 @run_only_rank0
-def export_allocation_stats(rank_allocation, filename):
+def export_allocation_stats(rank_allocation, filename, metype_memory_usage, memory_per_metype_file):
```
ferdonline, line 201: I would separate the two for a simple reason: An allocation can be generated anytime from `metype_memory_usage` and `memory_per_metype_file`, so we may want to store the parts independently

neurodamus/node.py
st4rl3ss: I resorted to using the `<` since the error can be a bit annoying on the user side if the number if ranks is not exactly the same. Would be cool to be able to redistribute on the fly if the number of ranks is not exactly the same though.

I'll revert to the old `!=` for the moment, in any case, then we can add this at a later stage.

neurodamus/utils/memory.py
st4rl3ss: Makes sense, will split (also rename a few variables because like this it's extremely unclear).

ferdonline (approved):

PR #180, opened 2024-07-03
[BBPBGLIB-975] Improved "signal" shot noise parameterization

seirios:
**NOTE: This is a breaking change for simulations using RelativeShotNoise or AbsoluteShotNoise stimuli**

The current "signal" shot noise parameterization (RelativeShotNoise and AbsoluteShotNoise stims) uses AmpCV together with the signal mean and standard deviation to derive the actual rate, amplitude mean and amplitude variance of the shot noise process. However, upon further analysis of the math, I realized that the AmpCV parameter is badly defined and does not represent a distinct property of the generated signal. This was further confirmed by observing the absence of an effect when changing this parameter to very different values, e.g., from 0.1 to 10.

Since the "signal" parameterization of shot noise already uses the mean and standard deviation (first two moments) of the signal, the natural extension was to consider the third moment as well, introducing a parameter associated to the skewness of the generated signal. For this particular type of shot noise (with bi-exponential shots and gamma-distributed amplitudes), there is a restricted range of possible skewness values for a given mean and standard deviation. This way, the parameter introduced is a "relative skewness" that goes from 0 to 1, with 0 representing the lowest and 1 the highest possible skewness for the generated signal, at the given mean and standard deviation.

The configuration parameter `AmpCV` (required) is replaced by `RelativeSkew` (optional, with default of 0.5). The inner workings of parameter derivation for RelativeShotNoise and AbsoluteShotNoise are modified to accommodate the new parameter.


neurodamus/stimulus_manager.py
```diff
@@ -422,16 +438,15 @@ def parse_check_stim_parameters(self, stim_info: dict):
         if self.sd <= 0:
             raise Exception("%s stdev must be positive" % self.__class__.__name__)
 
-        # coefficient of variation of shot amplitudes [1]
-        cv = float(stim_info["AmpCV"])
-        if cv <= 0:
-            raise Exception("%s amplitude CV must be positive" % self.__class__.__name__)
-        self.cv_square = cv * cv
+        # relative skewness of signal as a [0,1] fraction [1]
+        self.rel_skew = float(stim_info.get("RelativeSkew", 0.5))
```
WeinaJi, line 442: Hi @seirios , have you added the corresponding new parameter in the SONATA spec?

seirios: I have now :)

PR #185, opened 2024-07-24
Fix multiple issues with Memory Load Balance for MMB run

st4rl3ss:
## Context
As detailed in [https://bbpteam.epfl.ch/project/issues/browse/BBPBGLIB-1152](https://bbpteam.epfl.ch/project/issues/browse/BBPBGLIB-1152), multiple issues came up when trying to use the `neurodamus` memory load balance to run the MMB circuit.

In particular, the issues were a `list index out of range` error (that was caused by ranks/cycle sometimes not having any gids assigned) and an `Unpickling` error that is still misterious in its exact mechanics but that was seemingly caused by a corrupted broadcast of the allocation data.

## Scope
This MR implements a huge reworking of the memory load balance workflow:
-  the `distribute_cells` function is now a subroutine of `distribute_cells_with_validation`. This new function implements directly a check to make sure all ranks/cycle have at least one gid assigned and will automatically rework the distribution with smaller batches in case the condition is not respected. Furthermore, it integrates a dynamic batch size calculation in order to speedup the distribution.
- the `import_allocation_file` has been completely reworked to integrate both a caching mechanism (to avoid reloading the allocation file in every cycle) but also now return only the allocation table for the appropriate rank.
- the load balance memory workflow now correctly supports multi-population circuit
- disable spike compression when `neurodamus` is run in Load Balance Memory mode

## Testing
Due to the nature of the changes, and to standardize them, the test have been completely reworked with the use of fixture and lots of improvements to stability Furthermore, the test to distribute cells in multi-pop and multi cycles has been reworked to use mock object, since the inherent oscillation in memory measurements could lead to different results and distributions.

## Review

### DIFF
--- a/neurodamus/node.py
+++ b/neurodamus/node.py
@@ -451,6 +451,8 @@ class Node:
         else:
             loader_opts = {}
 
+        loader_opts["cycle_i"] = self._cycle_i
+
         # Check dynamic attributes required before loading cells
         SimConfig.check_cell_requirements(self.target_manager)
 
--- a/neurodamus/node.py
+++ b/neurodamus/node.py
@@ -377,7 +377,11 @@ class Node:
             logging.info("Load Balancing ENABLED. Mode: Memory")
             filename = f"allocation_r{MPI.size}_c{SimConfig.modelbuilding_steps}.pkl.gz"
 
-            if ospath.exists(filename):
+            file_exists = ospath.exists(filename)
+            MPI.barrier()
+
+            if file_exists:
+                logging.warning("calling import_allocation_stats from compute_load_balance")
                 alloc = import_allocation_stats(filename, self._cycle_i)
             else:
                 logging.warning("Allocation file not found. Generating on-the-fly.")
@@ -387,6 +391,7 @@ class Node:
                 cell_distributor.load_nodes(None, loader_opts={"load_mode": "load_nodes_metype",
                                                                "dry_run_stats": self._dry_run_stats}
                                             )
+                logging.warning("calling distribute_cells_with_validation from compute_load_balance")
                 alloc, _, _ = self._dry_run_stats.distribute_cells_with_validation(
                     MPI.size,
                     SimConfig.modelbuilding_steps,
@@ -505,6 +510,7 @@ class Node:
         if SimConfig.dry_run:
             self._dry_run_stats.collect_all_mpi()
             self._dry_run_stats.export_cell_memory_usage()
+            logging.warning("calling estimate_cell_memory from create_cells")
             self._dry_run_stats.estimate_cell_memory()
 
         # Final bits after we have all cell managers
@@ -1807,6 +1813,7 @@ class Neurodamus(Node):
             ranks = self._dry_run_stats.get_num_target_ranks(SimConfig.num_target_ranks)
             self._dry_run_stats.collect_all_mpi()
             try:
+                logging.warning("Calling distribute_cells_with_validation from .run")
                 self._dry_run_stats.distribute_cells_with_validation(ranks,
                                                                      SimConfig.modelbuilding_steps)
             except RuntimeError as e:
--- a/neurodamus/node.py
+++ b/neurodamus/node.py
@@ -387,7 +387,7 @@ class Node:
                 cell_distributor.load_nodes(None, loader_opts={"load_mode": "load_nodes_metype",
                                                                "dry_run_stats": self._dry_run_stats}
                                             )
-                alloc, _, _ = self._dry_run_stats.distribute_cells(
+                alloc, _, _ = self._dry_run_stats.distribute_cells_with_validation(
                     MPI.size,
                     SimConfig.modelbuilding_steps,
                     DryRunStats._MEMORY_USAGE_PER_METYPE_FILENAME
@@ -1804,7 +1804,8 @@ class Neurodamus(Node):
             self._dry_run_stats.display_node_suggestions()
             ranks = self._dry_run_stats.get_num_target_ranks(SimConfig.num_target_ranks)
             self._dry_run_stats.collect_all_mpi()
-            self._dry_run_stats.distribute_cells(ranks, SimConfig.modelbuilding_steps)
+            self._dry_run_stats.distribute_cells_with_validation(ranks,
+                                                                 SimConfig.modelbuilding_steps)
             return
         if not SimConfig.simulate_model:
             self.sim_init()
--- a/neurodamus/node.py
+++ b/neurodamus/node.py
@@ -398,19 +398,6 @@ class Node:
             for pop, ranks in alloc.items():
                 for rank, gids in ranks.items():
                     logging.debug(f"Population: {pop}, Rank: {rank}, Number of GIDs: {len(gids)}")
-            if MPI.rank == 0:
-                unique_ranks = set(
-                    rank[0] if isinstance(rank, tuple) else rank
-                    for pop in alloc.values()
-                    for rank in pop.keys()
-                )
-                logging.debug("Unique ranks in allocation file: %s", len(unique_ranks))
-                if MPI.size != len(unique_ranks):
-                    raise ConfigurationError(
-                        "The number of ranks in the allocation file is different from the number "
-                        "of ranks in the current run. The allocation file was created with a "
-                        "different number of ranks."
-                    )
             return alloc
 
         # Build load balancer as per requested options
--- a/neurodamus/node.py
+++ b/neurodamus/node.py
@@ -397,7 +397,6 @@ class Node:
             for pop, ranks in alloc.items():
                 for rank, gids in ranks.items():
                     logging.debug(f"Population: {pop}, Rank: {rank}, Number of GIDs: {len(gids)}")
-                    logging.debug(f"First 5 GIDs: {gids[:5]}")
             return alloc
 
         # Build load balancer as per requested options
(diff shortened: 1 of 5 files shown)


neurodamus/utils/memory.py
```diff
@@ -568,3 +578,90 @@ def validate_inputs_distribute(self, num_ranks, batch_size):
...
+        Returns:
+            Tuple[dict, dict, dict]: Returns the same as distribute_cells once a valid distribution
+                                     is found.
+        """
+        if not cycles:
+            cycles = 1
+
+        valid_distribution = False
+
+        def calculate_total_elements_per_population(self):
```
jorblancoa, line 629: nitpick:
I would name inner functions starting with _
_calculate_total_elements_per_population

neurodamus/utils/memory.py
```diff
@@ -568,3 +578,90 @@ def validate_inputs_distribute(self, num_ranks, batch_size):
...
+
+        Args:
+            num_ranks (int): The number of ranks.
+            cycles (int): The number of cycles to distribute cells over.
+            metype_file (str): The path to a JSON file containing memory usage for each METype.
+
+        Returns:
+            Tuple[dict, dict, dict]: Returns the same as distribute_cells once a valid distribution
+                                     is found.
+        """
```
jorblancoa, line 623: Maybe some logging would be nice at the beginning of the function?
Something like
logging.info(f"Starting distribution with {num_ranks} ranks and {cycles} cycles.")

neurodamus/utils/memory.py
```diff
@@ -568,3 +578,90 @@ def validate_inputs_distribute(self, num_ranks, batch_size):
...
+                total_elements_per_population[population] = total_elements
+            return total_elements_per_population
+
+        total_cells_per_population = calculate_total_elements_per_population(self)
+        average_cells_per_population = {
+            population: total / (num_ranks * cycles)
+            for population, total in total_cells_per_population.items()}
+
+        batch_size = {
+            population: max(1, int(average / 10))
```
jorblancoa, line 642: maybe im missing something, but where is the 10 coming from? is an arbitrary number? in that case maybe it should be defined as a constant?

neurodamus/utils/memory.py
st4rl3ss: This makes sense, I will adeguate the inner functions names.

neurodamus/utils/memory.py
st4rl3ss: 10 is just an arbitrary number in this sense. The idea is to allow the complete assignment of all gids in ten steps at most, since we've seen that it doesn't make a big difference in terms of balance.

Will definitely include a `const` for this.

neurodamus/utils/memory.py
```diff
@@ -568,3 +577,91 @@ def validate_inputs_distribute(self, num_ranks, batch_size):
...
+        while not valid_distribution and any(size > 0 for size in batch_size.values()):
+            bucket_allocation, bucket_memory, metype_memory_usage = self.distribute_cells(
+                num_ranks, cycles, metype_file, batch_size=batch_size
+            )
+            valid_distribution = self.check_all_buckets_have_gids(bucket_allocation,
+                                                                  num_ranks,
+                                                                  cycles)
+            if not valid_distribution:
+                batch_size = {population: max(0, size - 1)
+                              for population, size in batch_size.items()}
```
WeinaJi, line 654: It checks that every rank should contain at least 1 gid from every population, am I reading it correctly? If in the multiple-population case we have population A with a lot of cells, population B with a few cells, to distribute pop B evenly, we also need to reduce the batch size for pop A and redistribute pop A each time?

WeinaJi (commented):
Overall, it looks good to me. Just 2 questions.

neurodamus/node.py
```diff
@@ -377,37 +376,27 @@ def compute_load_balance(self):
             logging.info("Load Balancing ENABLED. Mode: Memory")
             filename = f"allocation_r{MPI.size}_c{SimConfig.modelbuilding_steps}.pkl.gz"
 
-            if ospath.exists(filename):
-                alloc = import_allocation_stats(filename, self._cycle_i)
+            file_exists = ospath.exists(filename)
+            MPI.barrier()
```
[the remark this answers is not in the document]
st4rl3ss: The barrier there was put to avoid a race condition. In that part of the code we are checking if the allocation file is there or not. If the file is not there it gets created just a few lines later. We could be in a condition where some of the ranks are still checking for the existence of the file while others are already creating it, so we could have some ranks passing the check because some of the other ranks on the same run have already created the file.

The race condition never happened, actually, but I've introduced it as a safety measure. If you think of any good reason why it shouldn't be there I can remove it, otherwise it's just a way to be sure the race condition never happens at no performance cost.

neurodamus/utils/memory.py
st4rl3ss: Well spotted Weina! 
To address this, I can modify the logic to reduce the batch size only for the populations that are causing the imbalance (i.e., populations with fewer cells). This way, I can avoid unnecessarily redistributing populations with many cells.

Thanks a lot for the comment!

st4rl3ss: @WeinaJi I've fixed the issue with the latest commit. Please give it a look when you have time.

neurodamus/utils/memory.py
```diff
@@ -568,3 +577,95 @@ def validate_inputs_distribute(self, num_ranks, batch_size):
...
+        logging.info("Distributing cells across %d ranks and %d cycles", num_ranks, cycles)
+
+        valid_distribution = False
+
+        def _calculate_total_elements_per_population(self):
+            total_elements_per_population = {}
+            for population, metypes in self.pop_metype_gids.items():
+                total_elements = sum(len(gids) for gids in metypes.values())
+                total_elements_per_population[population] = total_elements
+            return total_elements_per_population
```
ferdonline, line 633:
```suggestion
            return {
                population: sum(len(gids) for gids in metypes.values())
                for population, metypes in self.pop_metype_gids.items()
            }
```

neurodamus/utils/memory.py
```diff
@@ -568,3 +577,95 @@ def validate_inputs_distribute(self, num_ranks, batch_size):
...
+
+        total_cells_per_population = _calculate_total_elements_per_population(self)
+        average_cells_per_population = {
+            population: total / (num_ranks * cycles)
+            for population, total in total_cells_per_population.items()}
```
ferdonline, line 638: nitpick but IMO slightly more readable
```suggestion
            for population, total in total_cells_per_population.items()
        }
```

neurodamus/utils/memory.py
```diff
@@ -568,3 +577,95 @@ def validate_inputs_distribute(self, num_ranks, batch_size):
...
+
+        batch_size = {
+            population: max(1, int(average / MAX_ALLOCATION_STEPS))
+            for population, average in average_cells_per_population.items()}
+
+        while not valid_distribution and any(size > 0 for size in batch_size.values()):
```
ferdonline, line 644: It would be useful some description about this distribution loop logic as it's getting complex.
I understand that we try the distribution with increasingly smaller batches per population until the distribution is valid. 
But that means in the bad case it will be slower as we will make N distributions, when the Nth distribution is with a single element per batch. Also it will redo the work for all populations, even if some distributed correctly.
I thought we wanted to avoid the loop and make a single distribution with the adjusted batch size, where the last batch contains the remainder elements. Example with 92 cells and 1 rank we should get batch_size=10, where the last batch will be partially filled with the remaining two cells. Are there any constraints preventing us doing so?

tests/integration-e2e/test_dry_run_workflow.py
```diff
@@ -45,131 +76,148 @@ def test_dry_run_workflow(USECASE3):
...
-    nd = Neurodamus(
-        str(USECASE3 / "simulation_sonata.json"),
-        lb_mode="Memory",
-    )
 
-    nd.run()
-
-
-def test_dry_run_workflow_multi():
+@pytest.mark.parametrize("neurodamus_instance", [
```
ferdonline, line 111: nice!

neurodamus/utils/memory.py
```diff
@@ -568,3 +577,95 @@ def validate_inputs_distribute(self, num_ranks, batch_size):
...
+                total_elements_per_population[population] = total_elements
+            return total_elements_per_population
+
+        total_cells_per_population = _calculate_total_elements_per_population(self)
+        average_cells_per_population = {
+            population: total / (num_ranks * cycles)
+            for population, total in total_cells_per_population.items()}
+
+        batch_size = {
+            population: max(1, int(average / MAX_ALLOCATION_STEPS))
```
ferdonline, line 641:
```suggestion
            population: max(1, math.ceil(average / MAX_ALLOCATION_STEPS))
```

neurodamus/utils/memory.py
st4rl3ss: Thank you for your feedback!

As you said, the loop attempts to distribute cells with decreasing batch sizes until a valid distribution is found, which can be inefficient.

Your suggestion to perform a single distribution attempt with an adjusted batch size, where the last batch contains the remaining elements, makes sense and could improve efficiency. I have massively simplified the `distribute_cells_with_validation` function to implement this approach. The new implementation calculates the batch size based on the average number of cells per population and performs a single distribution attempt. If the distribution is not valid, it raises an error, prompting the user to adjust the number of ranks or cycles.

This approach should reduce the number of distribution attempts and improve overall performance. Please review the revised implementation and let me know if you have any further suggestions or concerns.

ferdonline (approved):
LGTM. Before merging it might be worth ensure it works with the circuits prepared for dry-run

PR #187, opened 2024-08-09
[BBPBGLIB-975] Improved "signal" shot noise parameterization

seirios:
**NOTE: This is a breaking change for simulations using RelativeShotNoise or AbsoluteShotNoise stimuli.**

The current "signal" shot noise parameterization (RelativeShotNoise and AbsoluteShotNoise stims) uses AmpCV together with the signal mean and standard deviation to derive the actual rate, amplitude mean and amplitude variance of the shot noise process. However, upon further analysis of the math, I realized that the AmpCV parameter is badly defined and does not represent a distinct property of the generated signal. This was further confirmed by observing the absence of an effect when changing this parameter to very different values, e.g., from 0.1 to 10.

Since the "signal" parameterization of shot noise already uses the mean and standard deviation (first two moments) of the signal, the natural extension was to consider the third moment as well, introducing a parameter associated to the skewness of the generated signal. For this particular type of shot noise (with bi-exponential shots and gamma-distributed amplitudes), there is a restricted range of possible skewness values for a given mean and standard deviation. This way, the parameter introduced is a "relative skewness" that goes from 0 to 1, with 0 representing the lowest and 1 the highest possible skewness for the generated signal, at the given mean and standard deviation.

The configuration parameter `AmpCV` (required) is replaced by `RelativeSkew` (optional, with default of 0.5). The inner workings of parameter derivation for RelativeShotNoise and AbsoluteShotNoise are modified to accommodate the new parameter.

### DIFF
--- a/neurodamus/stimulus_manager.py
+++ b/neurodamus/stimulus_manager.py
@@ -330,9 +330,9 @@ class ShotNoise(BaseStim):
         # nothing to do
         pass
 
-    def params_from_mean_var(self, mean, var):
+    def params_from_mean_sd(self, mean, sd):
         """
-        Compute bi-exponential shot noise parameters from desired mean and variance of signal.
+        Compute bi-exponential shot noise parameters from desired mean and std. dev. of signal.
 
         Analytical result derived from a generalization of Campbell's theorem present in
         Rice, S.O., "Mathematical Analysis of Random Noise", BSTJ 23, 3 Jul 1944.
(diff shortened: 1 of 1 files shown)


neurodamus/stimulus_manager.py
```diff
@@ -342,12 +342,30 @@ def params_from_mean_var(self, mean, var):
...
+
+        # cumulants
+        lambda2_1 = sd ** 2 / mean                     # lambda2 over lambda1
+        lambda3_2 = sd * skew                          # lambda3 over lambda2
+        theta1pk = 2 / (A * Xi) * lambda2_1            # = (1 + k) * theta
+        theta2pk = (3 * A) / (4 * B * Xi) * lambda3_2  # = (2 + k) * theta
+
+        # derived parameters
+        self.amp_mean = 2 * theta1pk - theta2pk               # mean amplitude [nA or uS]
+        self.amp_var = self.amp_mean * (theta2pk - theta1pk)  # variance of amplitude [nA^2 or uS^2]
```
matz-e, line 366: Why A _or_ S? Don't you need a voltage to convert between these two units?

ferdonline (approved):
Technically it's fine for me. The scientific correctness I can't say much, maybe a unit test could make sense, but leave that entirely to Sirio. If he already validated then I'm fine.

neurodamus/stimulus_manager.py
seirios: Hi @matz-e . It's one or the other, so no conversion is necessary. When the signal is injected as a Current, it has units of A, and when injected as a Conductance it has units of S. :)

seirios: Of course, when it's injected as a conductance, there is a voltage value involved (the holding voltage of the voltage clamp used to implement the dynamic clamp) to convert to a membrane current.

matz-e: Thank you for clarifying!

PR #190, opened 2024-08-14
Crash test mode

ferdonline:
## Context

Some simulations are pretty heavy, requiring a lot of nodes and time while, unfortunately, some problems only manifest in such very large scenarios. 

In order to improve the experience when testing new circuits, features or debugging a failure we introduce a new "Crash test" mode.

The new mode runs a simplified version of the circuit:
 - Cells feature a single section
 - Connections (between 2 cells) feature a single synapse (to the only section)

Therefore, while results are meaningless, the vast majority of Neurodamus internal machinery runs very quickly, allowing debug sessions to be orders of magnitude faster.

## Scope

 - Added new CLI option `--crash-test`
 - Defined a new full-python cell featuring a single section, with all required attributes
 - Implement a new `create_single()` function in Connection 

## Testing

Ran `sonataconf-sscx-O1` blueconfigs test (1000 cells) in a **single** rank: 
```
real	5m37.478s
user	3m44.838s
sys	0m3.816s
```
including 80 seconds for simulating 100 ms

## Review

### DIFF
--- a/neurodamus/metype.py
+++ b/neurodamus/metype.py
@@ -208,6 +208,32 @@ class EmptyCell(BaseCell):
         self.gid = gid
 
 
+class PointCell:
+    """Class representing a minimal, single section cell for dry-runs"""
+
+    def __init__(self, gid, cell_info, _circuit_conf):
+        self.gid = gid
+        self.raw_gid = None
+        self.soma = [Nd.Section(name="soma[0]", cell=self)]
+        self.exc_mini_frequency = float(cell_info.exc_mini_frequency)
+        self.inh_mini_frequency = float(cell_info.inh_mini_frequency)
+        self.synHelperList = []
+        self.synlist = []
+
+    CellRef = property(lambda self: self)
+    CCell = property(lambda self: self)
+    nSecAll = property(lambda _self: 1)
+    all = property(lambda self: [self.soma])
+    input_resistance = property(lambda _self: 1)
+
+    def connect2target(self, target_pp=None):
+        soma_sec = self.soma[0]
+        return Nd.NetCon(soma_sec(1)._ref_v, target_pp, sec=soma_sec)
+
+    def re_init_rng(self, ion_seed):
+        pass
+
+
 # Metadata
 # --------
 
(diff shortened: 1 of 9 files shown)


tests/integration-e2e/test_dry_run_workflow.py
WeinaJi, line 1: I guess `crash-test=True` will be used in one of the tests here?

ferdonline: The tests we have here are small and won't benefit much.
With the latest version we can "run" 200k cells in a single machine, so I think it would be nicer to have such a test in BlueConfigs

ferdonline: @WeinaJi I ended up adding an integration test here which instantiates 18k cells in this mode. In the following PR I will extend the test for synapses, since here it's still a bit slow

neurodamus/metype.py
```diff
@@ -208,6 +208,32 @@ def __init__(self, gid, cell):
...
+        self.exc_mini_frequency = float(cell_info.exc_mini_frequency)
+        self.inh_mini_frequency = float(cell_info.inh_mini_frequency)
+        self.synHelperList = []
+        self.synlist = []
+
+    CellRef = property(lambda self: self)
+    CCell = property(lambda self: self)
+    nSecAll = property(lambda _self: 1)
+    all = property(lambda self: self.soma)
+    input_resistance = property(lambda _self: 1)
```
WeinaJi, line 227: Is `input_resistance` set for stimulus? If so, do we also need to set `threshold_current` and `holding_current` ?

ferdonline: You have a point to have this API more complete. I clearly did the minimal changes only.
`threshold_current` and `holding_current` don't seem to be cells properties but rather of `METypeItem` metadata objects. However cell objects seem to commonly rely on `getThreshold` function. Let me add the common functions from `METype` then

WeinaJi (approved):

WeinaJi:
LGTM. Thanks for implementing this new feature.

PR #191, opened 2024-08-15
Crash-test mode: Speed up synapse loading

ferdonline:
## Context

Speed up crash test mode to avoid loading as much data

## Scope

 - `SonataReader` to quickly find out cells for which we have no data and mark them with an empty dict
 - For cells which have some data, in crash-test mode restrict the fetching edge index to only one (the first) per target cell. This operating could be done exclusively with numpy indexing operations, and is therefore fast.

### DIFF
--- a/neurodamus/connection_manager.py
+++ b/neurodamus/connection_manager.py
@@ -638,6 +638,8 @@ class ConnectionManagerBase(object):
             gids: Use given gids, instead of the circuit target cells. Default: None
             show_progress: Display a progress bar as tgids are processed
         """
+        CRASH_TEST_MODE_MAX_CONNS_CELL = 3
+        crash_test_mode = SimConfig.cli_options.crash_test
         if src_target and src_target.is_void() or dst_target and dst_target.is_void():
             return
 
@@ -654,7 +656,7 @@ class ConnectionManagerBase(object):
         sgid_offset, tgid_offset = self.get_population_offsets()
 
         self._synapse_reader.configure_override(mod_override)
-        self._synapse_reader.preload_data(gids)
+        self._synapse_reader.preload_data(gids, minimal_mode=crash_test_mode)
         extra_fields = {}  # Without extra fields, reuse this object
 
         # NOTE: This routine is quite critical, sitting at the core of synapse processing
@@ -663,12 +665,16 @@ class ConnectionManagerBase(object):
         # For each tgid we obtain the synapse parameters as a record array. We then split it,
         # without copying, yielding ranges (views) of it.
 
-        gids = ProgressBar.iter(gids) if show_progress else gids
+        gids = ProgressBar.iter(gids, name="Loading") if show_progress else gids
 
         for base_tgid in gids:
             tgid = base_tgid + tgid_offset
             syns_params = self._synapse_reader.get_synapse_parameters(base_tgid)
             logging.debug("GID %d Syn count: %d", tgid, len(syns_params))
+
+            if crash_test_mode:
+                # In large simulations some cells may have thousands of connections. Limit it!
+                syns_params = syns_params[:CRASH_TEST_MODE_MAX_CONNS_CELL]
 
             if self._load_offsets:
                 syn_index = self._synapse_reader.get_property(base_tgid, "synapse_index")
--- a/neurodamus/connection_manager.py
+++ b/neurodamus/connection_manager.py
@@ -628,7 +628,7 @@ class ConnectionManagerBase(object):
                 log_all(logging.DEBUG, "Source GIDs for debug cell: %s", self.yielded_src_gids)
 
     # -
-    def _iterate_conn_params(self, src_target, dst_target, gids=None, show_progress=False,
+    def _iterate_conn_params(self, src_target, dst_target, gids=None, show_progress=None,
                              mod_override=None):
         """A generator which loads synapse data and yields tuples(sgid, tgid, synapses)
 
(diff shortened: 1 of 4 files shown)


neurodamus/connection_manager.py
```diff
@@ -638,6 +638,7 @@ def _iterate_conn_params(self, src_target, dst_target, gids=None, show_progress=
             gids: Use given gids, instead of the circuit target cells. Default: None
             show_progress: Display a progress bar as tgids are processed
         """
+        crash_test_mode = SimConfig.cli_options.crash_test
```
matz-e, line 641: Could this be a dedicated `ConnectionManager`, inheriting from this class and overwriting some functionality instead?

ferdonline: We don't need any change in the `ConnectionManager` anymore, I can even move this flag to the reader. I did have that thought for the `SonataReader` though, where at some point changes were significant.
In the current latest implementation I reduced the changes to a couple of conditional blocks. Maybe that's acceptable

neurodamus/io/synapse_reader.py
```diff
@@ -358,10 +395,14 @@ def _load_params_custom(self, _populate, _read):
                 _populate("offset", _read("morpho_offset_segment_post"))
 
     def _load_synapse_parameters(self, gid):
-        if gid not in self._data:
-            self.preload_data([gid])
+        data = self._data.get(gid, False)
+        if data is False:  # strictly compare to False (not in _data)
```
WeinaJi, line 399: What is the advantage of this change of strictly compare to False?

neurodamus/io/synapse_reader.py
```diff
@@ -253,27 +255,57 @@ def get_property(self, gid, field_name):
...
+        different_gids_edge_i = np.diff(lookup_gids, prepend=np.nan).nonzero()[0]
+        needed_gids = sorted(lookup_gids[different_gids_edge_i])
+        for gid in (orig_needed_gids_set - set(needed_gids)):
+            self._data.setdefault(gid, self.EMPTY_DATA)
+
+        # In minimal mode read a single synapse (the first) of each target gid
+        if minimal_mode:
+            needed_edge_ids = libsonata.Selection(needed_edge_ids.flatten()[different_gids_edge_i])
+            lookup_gids = lookup_gids[different_gids_edge_i]
+            needed_gids = sorted(lookup_gids)
```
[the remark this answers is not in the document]
ferdonline: I'm replacing the variable, so that we exclude gids without edges. Lmk if you find the comment not enough

neurodamus/io/synapse_reader.py
ferdonline: to highlight `False` came from a missing key (line up). We must not use `if not data`, because `{}` (empty data) would match

neurodamus/io/synapse_reader.py
```diff
@@ -253,27 +255,57 @@ def get_property(self, gid, field_name):
...
+        Set minimal_mode to True to read a single synapse per connection
+        """
+        # Ensure we dont ask synapses for more than 50 target cells at once
+        CHUNK_SIZE = 50
+        if len(gids) <= CHUNK_SIZE:
+            return self._preload_data_chunk(gids, minimal_mode)
+
+        ranges = list(gen_ranges(len(gids), CHUNK_SIZE))
+        for start, end in ProgressBar.iter(ranges, name="Prefetching"):
+            self._preload_data_chunk(gids[start:end], minimal_mode)
```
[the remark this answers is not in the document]
ferdonline: yes. We never really had to care for extreme cases, but chunking is important to avoid exploding memory in case we have many cells in a single machine.

neurodamus/io/synapse_reader.py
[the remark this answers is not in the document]
ferdonline: Hum, good point, yes, in principle the needed gids doesn't change. Yeah, I think that will work if we delete that

neurodamus/io/synapse_reader.py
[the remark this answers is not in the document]
matz-e: Hm, I think we originally try to read synapses completely to minimize the amount of separate accesses to GPFS. Can we preserve that behavior?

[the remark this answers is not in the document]
ferdonline: We can adjust the CHUNK_SIZE to get as much data as possible. But I would like to cap memory consumption. With a CHUNK of 1000 cells, with 1000 syns/cell that's up to 1m rows per property at a time. Would it sounds reasonable?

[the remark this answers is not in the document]
ferdonline: Never mind, even if chunking is interesting we are doing collective IO, so the operation is not as trivial.
I added a comment and activated chunking only in crash-test

neurodamus/io/synapse_reader.py
ferdonline: The benefit is that now we do only one dict lookup, not two

matz-e: Does this optimization even matter when we are doing a lot of I/O? Even when we are reading a single synapse per GID, it seems like the less confusing, cleaner lookup would not be excessively punitive.

ferdonline: Once we get used to do this way it becomes kind of automatic, and the loss of readability I find minimal. How hard you want the old way?

WeinaJi (approved):
Looks good to me.
If @matz-e's comments are addressed, this PR can be merged.

neurodamus/io/synapse_reader.py
matz-e: That's better, thanks!

PR #194, opened 2024-08-28
Fix the Python 3.12 version inconsistency in GitHub actions cache 

WeinaJi:
## Context
GitHub actions/python-versions has recently released python `3.12.5` as the default one for python 3.12. But our py3.12 cache was built with python `3.12.4`.  This inconsistency leads to the `pip install` failure in the workflow test for python 3.12.  

## Scope
Specify python-version `3.12.5` in the workflow script.

## Testing
Existing test.

## Review

### DIFF
--- a/.github/workflows/simulation_test.yml
+++ b/.github/workflows/simulation_test.yml
@@ -201,6 +201,6 @@ jobs:
         mpirun -np 2 special -mpi -python $NEURODAMUS_PYTHON/init.py --configFile=simulation_sonata_coreneuron.json
         ls reporting_coreneuron/*.h5
 
-    - name: live debug session, comment out
-      if: failure()
-      uses: mxschmitt/action-tmate@v3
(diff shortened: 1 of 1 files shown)


WeinaJi:
Instead of this fix, we can remove the virtual_env cache in the workflow so that we don't need to worry about any actions/python-versions change in the future. The cost will be a few minutes more `pip install` process in each workflow. 
@ferdonline @jorblancoa , which way do you prefer?

ferdonline:
I think it's fine as it is! Next time, e.g. for 3.13, we will remember to choose to a specific Python release.

WeinaJi:
> I think it's fine as it is! Next time, e.g. for 3.13, we will remember to choose to a specific Python release.

Ok, could you please approve this PR  @ferdonline ? Thanks!

ferdonline (approved):

PR #198, opened 2024-10-02
Don't return early from target_gids, enable intersection

ferdonline:
## Review


jorblancoa (approved):

neurodamus/connection_manager.py
```diff
@@ -559,12 +559,14 @@ def connect_group(self, conn_source, conn_destination, synapse_type_restrict=Non
             mod_override (str): ModOverride given for this connection group
         """
         conn_kwargs = {}
-        pop = self._cur_population
+        conn_pop = self._cur_population
+        dst_pop_name = self._cell_manager.population_name
+        src_pop_name = self._src_cell_manager.population_name
```
WeinaJi, line 564: Just a nitpick:
Maybe we can set them as the attributes for `ConnectionManagerBase` ? I see that many places call them.

ferdonline: Yes, that is already in a follow up PR :)

WeinaJi (approved):

PR #200, opened 2024-10-17
ConnManager : Small readability improvements

ferdonline:
## Scope

Deduplicate and improve code readability (comments, vars...)

## Testing
No functional changes

## Review


neurodamus/connection_manager.py
```diff
@@ -840,13 +846,14 @@ def get_target_connections(self, src_target_name,
         if src_target and src_target.is_void() or dst_target.is_void():
             return
 
-        _, tgid_offset = self.get_population_offsets()
-        populations: List[ConnectionSet] = (conn_population,) if conn_population is not None \
+        tgid_offset = self.target_pop_offset
```
WeinaJi, line 849: Can we remove the func `get_population_offsets()` ?

neurodamus/connection_manager.py
```diff
@@ -328,8 +334,8 @@ def _open_synapse_file(self, synapse_file, pop_name):
 
     def _init_conn_population(self, src_pop_name, pop_id_override):
         if not src_pop_name:
-            src_pop_name = self._src_cell_manager.population_name
-        dst_pop_name = self._cell_manager.population_name
+            src_pop_name = self.src_node_population
+        dst_pop_name = self.dst_node_population
```
WeinaJi, line 338: There are also other places that use `conn_manager.src_cell_manager.population_name`, i.e. cell_distribution.py, gap_junction.py. We can also replace there.

neurodamus/connection_manager.py
ferdonline: good point. That was a reminiscent from the hoc targets... 😅

neurodamus/connection_manager.py
ferdonline: I fixed a few, but in other places we already have the node_manager, so we directly access the property instead of going all the way from the conn_manager

WeinaJi (approved):

jorblancoa (approved):

PR #210, opened 2024-11-25
[BBPBGLIB-1184] Fix CoreNEURON reports restore functionality

jorblancoa:
## Context
During CoreNEURON restore executions in multi-circuit simulations, reports fail to generate when circuits have multiple populations with different offsets. This occurs because the GIDs are saved in the report.conf without their proper population offsets during restore, as not all circuit population offsets are calculated at this stage. As a result, CoreNEURON ignores these reports due to GID mismatch with the original save execution.

To fix this, we now reuse the 'save' report.conf file and update only its tstop value, preserving the correct GID mappings.

## Review

### DIFF
--- a/neurodamus/node.py
+++ b/neurodamus/node.py
@@ -947,33 +947,33 @@ class Node:
 
         if SimConfig.restore_coreneuron:
             CoreConfig.update_tstop(rep_params.name, target_spec.name, rep_params.end)
-        else:
-            # for sonata config, compute target_type from user inputs
-            if "Sections" in rep_conf and "Compartments" in rep_conf:
-                def _compute_corenrn_target_type(section_type, compartment_type):
-                    sections = ["all", "soma", "axon", "dend", "apic"]
-                    compartments = ["center", "all"]
-                    if section_type not in sections:
-                        raise ConfigurationError(f"Report: invalid section type '{section_type}'")
-                    if compartment_type not in compartments:
-                        raise ConfigurationError(f"Report: invalid compartment type "
-                                                 f"{compartment_type}")
-                    if section_type == "all":  # for "all sections", support only target_type=0
-                        return 0
-                    # 0=Compartment, Section { 2=Soma, 3=Axon, 4=Dendrite, 5=Apical, 6=SomaAll ... }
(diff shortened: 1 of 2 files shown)


neurodamus/node.py
```diff
@@ -940,35 +942,38 @@ def _report_build_params(self, rep_name, rep_conf, target, pop_offsets_alias_pop
...
-        core_report_params = (
-            (os.path.basename(rep_conf.get("FileName", rep_name)),
-                target_spec.name, rep_params.rep_type, reporton_comma_separated)
-            + rep_params[3:5] + (target_type,) + rep_params[5:8]
-            + (target.get_gids(), SimConfig.corenrn_buff_size)
-        )
-        CoreConfig.write_report_config(*core_report_params)
+        if SimConfig.restore_coreneuron:
+            CoreConfig.update_tstop(rep_params.name, target_spec.name, rep_params.end)
+        else:
```
ferdonline, line 950: Very often I deal with the "exception case" (here restore) and return. Like that the general case does not get indented with IMO makes it feel "general" and helps readability. Also the diff would be smaller. But that's just  just another way, it's perfectly fine.

ferdonline (approved):

neurodamus/node.py
jorblancoa: You mean to have a return after the restore? makes sense yes...
I thought to have the other else as main if and the restore in the else but i didnt like it. The return early makes more sense

neurodamus/node.py
```diff
@@ -940,9 +942,13 @@ def _report_build_params(self, rep_name, rep_conf, target, pop_offsets_alias_pop
         )
 
     #
-    def _report_write_coreneuron_config(self, rep_name, rep_conf, target, rep_params):
+    def _report_write_coreneuron_config(self, rep_conf, target, rep_params):
         target_spec = TargetSpec(rep_conf["Target"])
 
+        if SimConfig.restore_coreneuron:
```
WeinaJi, line 948: Can we add a comment here addressing this solution is for the case of no change of reporting in restore.  For feature requiring different reports in restore, the work flow needs to be adapted.

ferdonline (approved):

WeinaJi (approved):

PR #212, opened 2024-11-26
Fix: Avoid extraneous newline

ferdonline:
## Scope

This PR fixes an issue in the output (rebalanced) files.dat in which the "number of files" header would be followed by a new line. This only occured when `--max-entries` was not being specified

## Testing
Tested with proj134/home/king/BBPP134-917/p_MultiCycle_Support


st4rl3ss (approved):
LGTM! Thanks a lot for fixing this!

PR #213, opened 2024-11-26
Rebalance tool: Fix record count. Enable using all files.dat entries

ferdonline:
## Scope

With this change we introduce the possibility of using all data from a files.dat even if the header specifies a different count.

The usage is `--max-entries=0`

We also update the header in the output file accordingly, i.e. independently of the original header, the output header specifies the number of records as written in the file.

## Testing

Tested with proj134/home/king/BBPP134-917/p_MultiCycle_Support

### DIFF
--- a/tools/rebalance-corenrn-data.py
+++ b/tools/rebalance-corenrn-data.py
@@ -42,7 +42,7 @@ def redistribute_files_dat(files_dat_file, n_buckets, max_entries=None, show_sta
     Read and process each entry from the dat file and distribute them into buckets.
 
     If max entries is not set (None) respect the number of entries from the header.
-    If user sets to `0` we use all entries form the file (disregard header)
+    If user sets to `0` we use all entries from the file (disregard header)
     """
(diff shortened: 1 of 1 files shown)


tools/rebalance-corenrn-data.py
```diff
@@ -39,6 +40,9 @@ def distribute_dat_to_bucket(dat_entry, size, buckets, bucket_sizes):
 def redistribute_files_dat(files_dat_file, n_buckets, max_entries=None, show_stats=False):
     """
     Read and process each entry from the dat file and distribute them into buckets.
+
+    If max entries is not set (None) respect the number of entries from the header.
+    If user sets to `0` we use all entries form the file (disregard header)
```
st4rl3ss, line 45:
```suggestion
    If user sets to `0` we use all entries from the file (disregard header)
```

st4rl3ss (approved):
Apart from the small typo, it LGTM!

st4rl3ss (approved):

WeinaJi (approved):

